PR #12, opened 2018-01-17
Fix error by setting DB ci collation

jochenschneider:
Test to reproduce case sensitive comparison error (which fails without the fix).


jochenschneider:
Deemed to risky, while be fixed in the OTA CE scripts.

PR #13, opened 2018-01-24
Add helm chart

alexhumphreys:


deploy/helm/device-registry/values.yaml
```diff
@@ -0,0 +1,43 @@
...
+service:
+  name: device-registry
+  type: ClusterIP
+  externalPort: 80
+  internalPort: 9001
+ingress:
+  enabled: false
+  # Used to create an Ingress record.
+  hosts:
+    - chart-example.local
```
alaa: can you change this to `device-registry.ota.local`

deploy/helm/device-registry/values.yaml
```diff
@@ -0,0 +1,43 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+replicaCount: 1
+image:
+  repository: advancedtelematic/device-registry
+  tag: 0.0.1-22-gc286097
+  pullPolicy: IfNotPresent
+service:
+  name: device-registry
+  type: ClusterIP
```
alaa: why it's using ClusterIP here?

deploy/helm/device-registry/values.yaml
```diff
@@ -0,0 +1,43 @@
...
+  externalPort: 80
+  internalPort: 9001
+ingress:
+  enabled: false
```
alaa: is this suppose to be `true` in our case?

alexhumphreys: it's an internal service, so I didn't think it'd have an ingress. but i can change it to true for testing if that's what we want.

deploy/helm/device-registry/templates/service.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "device-registry.fullname" . }}
```
alaa: seems like `.name` works as we expect.. not sure what's the point of `.fullname`

alexhumphreys:
Superseded by ota-charts repository.

PR #27, opened 2018-07-19
 Add endpoint to retrieve device group details.

houcros:


src/main/resources/application.conf
```diff
@@ -39,7 +39,7 @@ database {
   url = "jdbc:mariadb://localhost:3306/device_registry"
   url = ${?DB_URL}
   properties {
-    user = "sota"
+    user = "sota_test"
```
houcros: I saw this discrepancy while working on this. Should I commit it in another branch/ticket?

simao: you can fix it here.

PR #28, opened 2018-07-20
Guaranty a device deletion event is published

koshelev:
- Web API endpoint publishes an event to kafka
- background job subscribes to events and deletes the device

### DIFF
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/daemon/DeleteDeviceHandler.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/daemon/DeleteDeviceHandler.scala
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2018 HERE Technologies
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package com.advancedtelematic.ota.deviceregistry.daemon
+
+import akka.Done
+import akka.actor.ActorSystem
+import com.advancedtelematic.libats.messaging.MessageListener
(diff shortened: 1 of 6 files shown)


src/main/scala/com/advancedtelematic/ota/deviceregistry/daemon/DeleteDeviceHandler.scala
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) 2017 ATS Advanced Telematic Systems GmbH
```
jochenschneider: 2018

src/main/scala/com/advancedtelematic/ota/deviceregistry/daemon/DeleteDeviceHandler.scala
```diff
@@ -0,0 +1,36 @@
...
+) {
+
+  def start(): Unit = {
+    val listener = system.actorOf(MessageListener.props[DeleteDeviceRequest](system.settings.config, handle, metrics),
+                                  "delete-device-handler")
+    listener ! Subscribe
+  }
+
+  private[this] def handle(message: DeleteDeviceRequest): Future[Done] =
+    db.run(DeviceRepository.delete(message.namespace, message.uuid).map(_ => Done))
```
jochenschneider: I wonder about error handling for this mechanism. Would be good to at least have a debug log here for debugging.

koshelev: Flow will handle it - log and retry

jochenschneider: Ah cool. Where is that set up?

koshelev: Somewhere in `MessageListener`

PR #31, opened 2018-08-07
Bump libats

simao:
Move DeviceEventmessage to libats


simao:
@jerrytrieu  can I merge this?

PR #34, opened 2018-08-16
Fix health check routes

simao:


simao:
I messed up the health check with this change in a previous commit.

cc @alexhumphreys

PR #35, opened 2018-08-30
Smart groups: new filter for 'deviceId char at position'

houcros:


simao:
I added a commit

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/GroupExpressionParser.scala
```diff
@@ -91,7 +89,7 @@ object GroupExpressionParser {
     _ <- skipWhitespace
     _ <- token(string("is"))
     char <- letterOrDigit
-  } yield DeviceIdCharAt(char, pos - 1)
+  } yield DeviceIdCharAt(char.toLower, pos - 1)
```
simao: this should only be done at the evaluation, we should keep the user input here, but compare only case insensitive

houcros: Indeed. I've amended my last commit.

PR #39, opened 2018-09-14
Sort groups by name and creation time.

houcros:


src/test/scala/com/advancedtelematic/ota/deviceregistry/GroupRequests.scala
```diff
@@ -48,7 +50,12 @@ trait GroupRequests {
   def countDevicesInGroup(groupId: GroupId)(implicit ec: ExecutionContext): HttpRequest =
     Get(Resource.uri("device_groups", groupId.show, "count"))
 
-  def listGroups(): HttpRequest = Get(Resource.uri(groupsApi))
+  def listGroups(sortBy: Option[SortBy] = None, limit : Option[Long] = None): HttpRequest = {
+    val m = mutable.Map[String, String]()
```
koshelev: WDYT about
```scala
    val m = List("sortBy" -> sortBy, "limit" -> limit).foldLeft(Map.empty[String, String])((acc, x) => x match {
      case n -> Some(x) => acc.updated(n, x.toString)
      case _ -> None => acc
    })
    Get(Resource.uri(groupsApi).withQuery(Query(m)))
```

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Group.scala
```diff
@@ -75,3 +81,17 @@ object Group {
     io.circe.generic.semiauto.deriveEncoder[Group]
   }
 }
+
+object SortBy extends Enumeration {
+  type SortBy = Value
+  val NAME       = Value("name")
+  val CREATED_AT = Value("createdAt")
+
+  implicit class Sorting(sortBy: Value) {
```
koshelev: Please move `Sorting` into `db` package

src/test/scala/com/advancedtelematic/ota/deviceregistry/GroupRequests.scala
houcros: More functional, nice 👍 Updated.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Group.scala
houcros: Moved to `DbOps.`.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Group.scala
```diff
@@ -8,23 +8,28 @@
...
+import slick.lifted.ColumnOrdered
 
 case class Group(id: GroupId,
                  groupName: Name,
                  namespace: Namespace,
                  groupType: GroupType,
-                 expression: Option[GroupExpression] = None)
+                 expression: Option[GroupExpression] = None,
+                 createdAt: Instant,
+                 updatedAt: Instant)
```
simao: we don't need timestamps here, we only need them at schema level when sorting.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Group.scala
```diff
@@ -44,6 +49,7 @@ object Group {
 
   case class ValidName()
   type Name = Refined[String, ValidName]
+  implicit val nameOrdering: Ordering[Name] = (a, b) => a.value.toLowerCase compareTo b.value.toLowerCase
```
simao: `implicit val nameOrdering: Ordering[Name] = Ordering.by[Name, String](_.value.toLowerCase)`

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/GroupInfoRepository.scala
```diff
@@ -34,19 +38,20 @@ object GroupInfoRepository extends SlickJsonHelper with ColumnTypes {
     def namespace  = column[Namespace]("namespace")
     def groupType     = column[GroupType]("type")
     def expression = column[Option[GroupExpression]]("expression")
+    def createdAt = column[Instant]("created_at")
+    def updatedAt = column[Instant]("updated_at")
 
-    def * = (id, groupName, namespace, groupType, expression) <> ((Group.apply _).tupled, Group.unapply)
+    def * = (id, groupName, namespace, groupType, expression, createdAt, updatedAt) <> ((Group.apply _).tupled, Group.unapply)
```
simao: don't need it

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/DbOps.scala
```diff
@@ -0,0 +1,18 @@
...
+import slick.lifted.ColumnOrdered
+
+object DbOps {
+  implicit class Sorting(sortBy: SortBy) {
+    def groupSorting: GroupInfoTable => ColumnOrdered[_] = sortBy match {
+      case SortBy.NAME       => _.groupName.asc
+      case SortBy.CREATED_AT => _.createdAt.desc
+    }
+  }
+}
```
simao: I think the idiomatic way would be to have the following in scope:

```scala
  implicit def sortBySlickOrderedConversion(sortBy: SortBy): GroupInfoTable => slick.lifted.Ordered =
    sortBy match {
      case SortBy.NAME => table => table.groupName.asc
      case SortBy.CREATED_AT => _.createdAt.desc
    }
```

and then you can simply do:

```scala
groupInfos
      .filter(g => g.namespace === namespace)
      .paginateAndSortResult(sortBy, offset, limit)
```

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Group.scala
```diff
@@ -75,3 +81,9 @@ object Group {
     io.circe.generic.semiauto.deriveEncoder[Group]
   }
 }
+
+object SortBy extends Enumeration {
+  type SortBy = Value
+  val NAME       = Value("name")
+  val CREATED_AT = Value("createdAt")
+}
```
simao: Since we are doing pattern matching on these, you should consider making `SortBy` a `sealed trait`, it might require more boiler plate on the Unmashallers but it does give us a bit more type safety.

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/GroupInfoRepository.scala
```diff
@@ -34,19 +38,20 @@ object GroupInfoRepository extends SlickJsonHelper with ColumnTypes {
...
 
-  def list(namespace: Namespace, offset: Option[Long], limit: Option[Long])(
-      implicit ec: ExecutionContext
-  ): DBIO[PaginationResult[Group]] =
+  def list(namespace: Namespace, sortBy: SortBy, offset: Long, limit: Long)(implicit ec: ExecutionContext): DBIO[PaginationResult[Group]] =
     groupInfos
       .filter(g => g.namespace === namespace)
-      .paginateResult(offset.getOrElse(0L), limit.getOrElse(defaultLimit))
+      .sortBy(sortBy.groupSorting)
+      .paginateResult(offset, limit)
```
simao: see above

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/GroupInfoRepository.scala
```diff
@@ -57,14 +62,9 @@ object GroupInfoRepository extends SlickJsonHelper with ColumnTypes {
...
-             groupName: Name,
-             namespace: Namespace,
-             groupType: GroupType,
-             expression: Option[GroupExpression])(
-      implicit ec: ExecutionContext
-  ): DBIO[GroupId] =
-    (groupInfos += data.Group(id, groupName, namespace, groupType, expression))
+  def create(id: GroupId, groupName: Name, namespace: Namespace, groupType: GroupType, expression: Option[GroupExpression])
+            (implicit ec: ExecutionContext): DBIO[GroupId] =
+    (groupInfos += data.Group(id, groupName, namespace, groupType, expression, Instant.now(), Instant.now()))
```
simao: will have different create and updated dates.

src/test/scala/com/advancedtelematic/ota/deviceregistry/GroupRequests.scala
```diff
@@ -48,7 +48,13 @@ trait GroupRequests {
   def countDevicesInGroup(groupId: GroupId)(implicit ec: ExecutionContext): HttpRequest =
     Get(Resource.uri("device_groups", groupId.show, "count"))
 
-  def listGroups(): HttpRequest = Get(Resource.uri(groupsApi))
+  def listGroups(sortBy: Option[SortBy] = None, limit : Option[Long] = None): HttpRequest = {
+    val m = List("sortBy" -> sortBy, "limit" -> limit).foldLeft(Map.empty[String, String])((acc, x) => x match {
+        case (k, Some(v)) => acc + (k -> v.toString)
+        case (_, None)    => acc
+    })
```
simao: `List("sortBy" -> sortBy, "limit" -> limit).collect { case (k, Some(v)) => k -> v.toString }.toMap`

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/DbOps.scala
houcros: I see that we only need the `createdAt` field at schema level, but how can I keep this `_.createdAt.desc` if we remove it from the model? Or what can I use instead?

simao: GroupInfo table will still have `createdAt`, so that isn't a problem. You just need to remove it from the `*` projection.

src/main/scala/com/advancedtelematic/ota/deviceregistry/GroupsResource.scala
```diff
@@ -42,8 +40,8 @@ class GroupsResource(
     }
   }
 
-  implicit val groupTypeParamUnmarshaller: Unmarshaller[String, GroupType] =
-    Unmarshaller.strict[String, GroupType](GroupType.withName)
+  implicit val groupTypeParamUnmarshaller: Unmarshaller[String, GroupType] = Unmarshaller.strict[String, GroupType](GroupType.withName)
+  implicit val sortByUnmarshaller: Unmarshaller[String, SortBy] = Unmarshaller.strict[String, SortBy](SortBy.withName)
```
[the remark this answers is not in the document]
houcros: Using an unmarshaller for cases now.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Group.scala
houcros: Removed.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Group.scala
houcros: Thanks, updated.

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/GroupInfoRepository.scala
houcros: Removed.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Group.scala
houcros: Changed for an abstract class with two concrete case classes.

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/GroupInfoRepository.scala
houcros: Updated.

src/test/scala/com/advancedtelematic/ota/deviceregistry/GroupRequests.scala
houcros: Even nicer 😄 Updated.

src/test/scala/com/advancedtelematic/ota/deviceregistry/GroupRequests.scala
```diff
@@ -48,16 +49,22 @@ trait GroupRequests {
   def countDevicesInGroup(groupId: GroupId)(implicit ec: ExecutionContext): HttpRequest =
     Get(Resource.uri("device_groups", groupId.show, "count"))
 
-  def listGroups(): HttpRequest = Get(Resource.uri(groupsApi))
+  def listGroups(sortBy: Option[SortBy] = None, limit : Option[Long] = None): HttpRequest = {
+    val m = List("sortBy" -> sortBy, "limit" -> limit).collect { case (k, Some(v: SortBy)) => k -> v.name }.toMap
```
simao: this wont actually work using a sealed trait though, it's filtering out `limit` a new `case` needs to be added to `collect`

src/test/scala/com/advancedtelematic/ota/deviceregistry/GroupsResourceSpec.scala
```diff
@@ -42,10 +44,40 @@ class GroupsResourceSpec extends FunSuite with ResourceSpec {
     }
   }
 
+  test("gets all existing groups sorted by name") {
+    val groupNames = Gen.listOfN(20, genGroupName).sample.get
+    val sortedGroupNames = groupNames.sorted
```
simao: I just noticed that the Ordering is only used in this test, in this line. This can just be rewritten as `val sortedGroupNames = groupNames.sortBy(_.value)` and then you can get rid of `Ordering[Name]`

simao:
And https://github.com/advancedtelematic/ota-device-registry/commit/fd4150fd113ccb284fbd611d5570c84f2f97a562

houcros:
@simao Thank you. I updated as per your comments.

PR #40, opened 2018-09-17
Search groups by name

houcros:
Based on `feat/OTA-623/sort-groups-updates-campaigns`. Merge after #39 .


houcros:
Moved to #42.

PR #41, opened 2018-09-19
Sort groups by name and creation time.

houcros:


src/test/scala/com/advancedtelematic/ota/deviceregistry/data/GroupGenerators.scala
```diff
@@ -19,16 +19,14 @@ trait GroupGenerators {
 
   val genGroupName: Gen[Group.Name] = for {
     strLen <- Gen.choose(2, 100)
-    name   <- Gen.listOfN[Char](strLen, Arbitrary.arbChar.arbitrary)
+    name   <- Gen.listOfN[Char](strLen, Gen.alphaNumChar)
```
houcros: I had to change this if we were to make any sense of sorting by group name. Hopefully this doesn't cause any regression (?)

simao: what do you mean? Why did you have to change it ?

houcros: We were generating group names with any Unicode chars. It's unclear to me how to assert they're sorted, hence very awkward to test. I don't think sorting by code-point is what we want.

simao: We do want to keep testing with unicode chars though. Maybe you could use a different generator for your test...

houcros: Yes, that was indeed my first approach. But that would imply that we accept any chars for the name, but sort only over ASCII chars right?
If we're fine with that I will just use another group name generator.

simao: that is not great but it's better than not testing with unicode at all

koshelev: If i understand it correctly, our API allows any valid UTF-8 character as an input. So I guess the test should have the same constraint

houcros: So: I added a new group name generator that it's used only for the sorting test.

src/test/scala/com/advancedtelematic/ota/deviceregistry/data/GroupGenerators.scala
```diff
@@ -22,13 +22,16 @@ trait GroupGenerators {
     name   <- Gen.listOfN[Char](strLen, Arbitrary.arbChar.arbitrary)
   } yield Refined.unsafeApply(name.mkString)
 
-  def genGroupInfo: Gen[Group] =
-    for {
-      name <- genGroupName
-    } yield Group(GroupId.generate(), name, defaultNs, GroupType.static, None)
+  val genAlphaNumGroupName: Gen[Group.Name] = for {
```
simao: I think would be better to have only one generator that accepts a Gen[Char] with `Arbitrary.arbChar.arbitrary` as the default, then you can pass `Gen.alphaNumChar` on your test

simao:
lgtm

PR #42, opened 2018-09-19
Search groups by name.

houcros:


houcros:
Based on #41.

houcros:
For OTA-622. I noticed we currently use regex to search devices.
Shall we do some "string contains" check instead of regex here? It seems that would still meet the ticket description.

simao:
Yeah that would be better

src/main/scala/com/advancedtelematic/ota/deviceregistry/GroupsResource.scala
```diff
@@ -86,8 +88,8 @@ class GroupsResource(namespaceExtractor: Directive1[AuthedNamespaceScope], devic
       (scope.post & entity(as[CreateGroup]) & pathEnd) { req =>
         createGroup(req.name, ns.namespace, req.groupType, req.expression)
       } ~
-      (scope.get & pathEnd & parameters(('offset.as[Long] ? 0L, 'limit.as[Long] ? 50L, 'sortBy.as[SortBy].?))) {
-        (offset, limit, sortBy) => listGroups(ns.namespace, offset, limit, sortBy.getOrElse(SortBy.Name))
+      (scope.get & pathEnd & parameters(('offset.as[Long] ? 0L, 'limit.as[Long] ? 50L, 'sortBy.as[SortBy].?, 'contains.as[String].?))) {
```
houcros: I'm not completely happy with this query param name, but I couldn't think of anything nicer.

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/GroupInfoRepository.scala
```diff
@@ -47,9 +47,13 @@ object GroupInfoRepository extends SlickJsonHelper with ColumnTypes {
 
   val groupInfos = TableQuery[GroupInfoTable]
 
-  def list(namespace: Namespace, offset: Long, limit: Long, sortBy: SortBy)(implicit ec: ExecutionContext): DBIO[PaginationResult[Group]] =
+  def list(namespace: Namespace, offset: Long, limit: Long, sortBy: SortBy, contains: Option[String])(implicit ec: ExecutionContext): DBIO[PaginationResult[Group]] =
     groupInfos
       .filter(_.namespace === namespace)
+      .withFilter(gi => contains match {
```
houcros: @simao I think there should be a way to extract this into libats, something similar to `com.advancedtelematic.libats.slick.db.SlickResultExtensions.QueryOps#maybeFilter`, but I couldn't quite manage. It would be nice because we'll do the same for campaigns and updates in campaigner. WDYT?

simao: Yes this could be added to libats, something like:

```scala
def maybeFilter2(f: E => Option[Rep[Boolean]]): Query[E, U, Seq] =
      query.withFilter { (e: E) =>
        f(e).getOrElse(true.bind)
      }
```

Or did you want to abstract the contains?

```scala
def maybeFilterContains(f: E => Rep[_], exp: Option[String]): Query[E, U, Seq] =
      query.withFilter { e: E =>
        exp match {
          case Some(s) if s.nonEmpty => f(e).mappedTo[String].like(s"%$s%")
          case _ => true.bind
        }
      }
```

src/main/scala/com/advancedtelematic/ota/deviceregistry/GroupsResource.scala
simao: What don't you like about it ?

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/GroupInfoRepository.scala
houcros: The contains, great I'll move it to libats 👍 I was close but not quite there 😄

src/main/scala/com/advancedtelematic/ota/deviceregistry/GroupsResource.scala
houcros: It kind of makes it seem we're searching all the group fields, but we're searching only in the name. `nameContains` is too verbose maybe, so we're fine with `contains`?

simao: I think `nameContains` is a good idea

houcros: `nameContains` it is then :)

PR #44, opened 2018-10-04
Create endpoint to count smart group candidates.

houcros:


houcros:
Once again I forgot to add the hat suffix. Will open a new PR.

simao:
I don't think we should add suffixes to branch names. This pollutes git history with something that does not belong in git history, /berlin will be in the merge commit.

I think we should have just a branch named something like deploy/berlin and push there, then it gets deployed to berlin.

wdyt @jerrytrieu  ?

jerrytrieu:
@simao  Yes, good point. Let's do that.

PR #46, opened 2018-10-05
Index all device events by type and correlation id

simao:


src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Codecs.scala
```diff
@@ -0,0 +1,8 @@
+package com.advancedtelematic.ota.deviceregistry.data
```
houcros: Could we not remove this class?

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
```diff
@@ -0,0 +1,41 @@
+package com.advancedtelematic.ota.deviceregistry.db
+
+import cats.syntax.either._
```
houcros: Some unused imports?

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Codecs.scala
simao: No, we are using it in a couple of places.

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventJournal.scala
```diff
@@ -47,20 +50,63 @@ object EventJournal {
...
     override def * =
-      (deviceUuid, eventId, eventTypeId, eventTypeVersion, deviceTime, receivedAt, event).shaped <> (toEvent _, fromEvent _)
+      (deviceUuid, eventId, eventTypeId, eventTypeVersion, deviceTime, receivedAt, event).shaped <> (toEvent, fromEvent)
   }
 
-  private[EventJournal] val events = TableQuery[EventJournalTable]
+  protected [db] val events = TableQuery[EventJournalTable]
+
+  class IndexedEventTable(tag: Tag) extends Table[IndexedEvent](tag, "IndexedEvents") {
+    def deviceId = column[DeviceUUID]("device_uuid")
```
houcros: Why not `deviceUuid` as in `EventJournalTable`?

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
simao: it's required for `asRight` and `asLeft`

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventJournal.scala
simao: I will change EventJournalTable in a separate pr.

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventJournal.scala
```diff
@@ -47,20 +50,63 @@ object EventJournal {
...
+    def eventId = column[String]("event_id")
+    def eventType = column[IndexedEventType]("event_type")
+    def correlationId = column[Option[CorrelationId]]("correlation_id")
+
+    def pk = primaryKey("indexed_event_pk", (deviceId, eventId))
```
houcros: Should we also enforce the FK restrictions here, or there's no advantage in that?

simao: The pk we have to enforce because slick uses that for insertOrUpdate, but for foreign keys slick doesn't use at all, so no advantage.

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
houcros: Oh sorry, I didn't mean that one in particular. I checkout out the branch and IntelliJ showed me other unused imports.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Codecs.scala
houcros: I found no usages 😕

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventJournal.scala
houcros: But wouldn't `deviceUuid` be more appropriate, so as to not to mix it it up with `deviceId`?

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
simao: fixed that

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Codecs.scala
simao: https://github.com/advancedtelematic/ota-device-registry/pull/46/files#diff-57691ae7c49407a9e875d9b30f8bcb4bR38 for example

simao: Sorry, actually the usages are on the next pr I have after this is merged, I will need this class.

houcros: Alright, I was starting to think I was blind 😄

src/main/resources/db/migration/V26__add_indexed_events.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IndexedEvents (
+  device_uuid char(36) COLLATE utf8_bin NOT NULL,
+  event_id char(36) NOT NULL,
+  correlation_id char(36) NOT NULL,
+  `event_type` enum('DownloadComplete','InstallationComplete') DEFAULT NULL,
```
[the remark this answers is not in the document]
simao: I will remove the enum from the database and use a string, but I think at code level we still should use some restriction and not accept any value, this means we'll still need some small code changes when adding a new state.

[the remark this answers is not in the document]
simao: updated

PR #47, opened 2018-10-08
Support smart groups expression 'deviceid position is not'.

houcros:


houcros:
I see what you mean, makes sense. I've borrowed from your commit :)

PR #50, opened 2018-10-12
Create endpoints to request devices by group type or ungrouped.

houcros:
Tested in *berlin* with this steps:

1. Create 3 new devices d1, d2, d3.
2. `GET /api/v1/devices/ungrouped` returns d1, d2, d3.
3. Create a new static group g1 and add d1 to g1.
5. `GET /api/v1/devices/ungrouped` returns d2, d3. `GET /api/v1/devices/grouped?groupType=static` returns d1.
6. Create a new dynamic group g2 so that it will only contain d2.
7. `GET /api/v1/devices/ungrouped` returns d3. `GET /api/v1/devices/grouped?groupType=dynamic` returns d2.


src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -183,6 +187,12 @@ class DevicesResource(
         (path("count") & parameter('expression.as[GroupExpression].?)) {
           case None      => complete(Errors.InvalidGroupExpression(""))
           case Some(exp) => countDynamicGroupCandidates(ns.namespace, exp)
+        } ~
+          (path("ungrouped") & parameters(('offset.as[Long].?, 'limit.as[Long].?))) { (offset, limit) =>
```
simao: ungrouped should still be a param and not part of the path, it is just a filter on the resource, it should not belong on the uri.

To do this you'll need to have a separate parameter for grouped=boolean groupType=type

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/DeviceRepository.scala
```diff
@@ -146,24 +147,38 @@ object DeviceRepository {
...
+
+      case Some(gt) => GroupInfoRepository.groupInfos
+        .filter(_.groupType === gt)
+        .join(GroupMemberRepository.groupMembers)
+        .on(_.id === _.groupId)
+        .join(DeviceRepository.devices)
+        .on(_._2.deviceUuid === _.uuid)
+        .map(_._2)
+        .paginateResult(offset.getOrElse(0L), limit.getOrElse[Long](defaultLimit).min(maxLimit))
+    }
```
simao: What are you trying to acomplish here? To return ungrouped devices when groupType is None, or return all grouped devices when groupType is not None ?

houcros: I want to return only ungrouped devices when groupType is None.

simao: I think that is a bit misleading. the method is called `searchByGrouping`, the caller would expect to only return grouped devices, and if groupType is None, then from all groups, but not ungrouped devices.

houcros: Right. I've refactored in two methods: one for grouped and one for ungrouped.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
houcros: I've updated using new query parameters.

src/test/scala/com/advancedtelematic/ota/deviceregistry/DeviceResourceSpec.scala
```diff
@@ -68,6 +67,19 @@ class DeviceResourceSpec extends ResourcePropSpec with ScalaFutures with Eventua
     }
   }
 
+  private def createGroupedAndUngroupedDevices(n: Int): Map[String, Seq[DeviceUUID]] = {
```
simao: we only ever use 12 anyway so lets get rid of `n` and just use the same number all the time. Then bellow just use constant numbers instead of divisions. This is quite hard to read

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/DeviceRepository.scala
```diff
@@ -146,28 +144,35 @@ object DeviceRepository {
...
-      case Some(re) => byNamespace.filter(d => regex(d.deviceName, re))
-      case None     => byNamespace
-    }
-    val grouped     = GroupMemberRepository.groupMembers.map(_.deviceUuid)
-    val byUngrouped = byOptionalRegex.filterNot(_.uuid in grouped)
-
-    byUngrouped.paginateAndSortResult(_.deviceName,
-                                      offset.getOrElse(0L),
-                                      limit.getOrElse[Long](defaultLimit).min(maxLimit))
+  def searchGrouped(ns: Namespace, groupType: Option[GroupType], offset: Option[Long], limit: Option[Long])
```
simao: This can be simplified a bit, checkout https://github.com/advancedtelematic/ota-device-registry/commit/033045aef0803038f889b3dbae38cde5989f354f

houcros:
Anything holding us from merging this?

simao:
lgtm

PR #52, opened 2018-10-12
Refactor pagination and groups generators in tests.

houcros:
* Unify the defaults for pagination in a single place in the DB.
* Refactor the groups generators for the tests.
* Other minor refactors.


src/main/scala/com/advancedtelematic/ota/deviceregistry/GroupMembership.scala
```diff
@@ -54,12 +52,8 @@ protected class StaticMembership(implicit db: Database, ec: ExecutionContext) ex
...
     db.run(GroupMemberRepository.removeGroupMember(group.id, deviceId))
 
-  def create(
-      groupId: GroupId,
-      name: Name,
-      namespace: Namespace
-  ): Future[GroupId] = db.run {
-    GroupInfoRepository.create(groupId, name, namespace, GroupType.static, expression = None)
+  def create(groupId: GroupId, name: Name, namespace: Namespace): Future[GroupId] = db.run {
+    GroupInfoRepository.create(groupId, name, namespace, GroupType.static, None)
```
simao: `GroupType.static, expression = None` is clearer.

PR #54, opened 2018-10-17
Revert the renaming of deviceUuid and deviceId.

houcros:


src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Device.scala
```diff
@@ -22,22 +22,22 @@ import eu.timepit.refined.api.{Refined, Validate}
 import io.circe.{Decoder, Encoder}
 
 final case class Device(namespace: Namespace,
-                        id: DeviceId,
-                        oemId: DeviceOemId,
-                        name: DeviceName,
+                        uuid: DeviceId,
```
simao: I think this can stay, afaik this is not going to the client or bus.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Device.scala
```diff
@@ -22,22 +22,22 @@ import eu.timepit.refined.api.{Refined, Validate}
...
+                        deviceName: DeviceName,
+                        deviceId: DeviceOemId,
                         deviceType: DeviceType = DeviceType.Other,
                         lastSeen: Option[Instant] = None,
                         createdAt: Instant,
                         activatedAt: Option[Instant] = None,
                         deviceStatus: DeviceStatus = NotSeen) {
 
-  def toResponse: DeviceT = DeviceT(oemId, name, deviceType)
+  def toResponse: DeviceT = DeviceT(deviceName, deviceId, deviceType)
```
simao: this method can go I don't think it's being used.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Device.scala
houcros: It goes to the FE when we return a list of devices. Though I doubt they use it...

simao: which endpoint is that?

houcros: `/api/v1/devices` where we can search devices.

simao: I see. well this is a bug then, this class should never leak out to the client, that is the whole point of DeviceT. Lets keep this for now then... unless they are not using it

houcros: Also most of the messages that go to the bus send the uuid, e.g. `final case class DeviceActivated(namespace: Namespace, uuid: DeviceId, at: Instant)`. If I don't revert this we have to change the consumers of the messages right?

houcros: Mmh I see, only if I had changed the property name in the previous PR.

simao: the only thing you cannot change is the attribute name, `uuid`, the type can change all you want, and actually on the other end the consumers already use `DeviceId` and not `Uuid`.

houcros: Got it. We're using `DeviceId` from libats everywhere after this PR.
About the bug, it seems FE does use the uuid, for all the endpoints like `/api/v1/devices/{uuid}/events`, `/api/v1/devices/{uuid}/packages`, etc. How would they get the uuid if we returned `DeviceT` from `/api/v1/devices`?

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Device.scala
houcros: Right, it's gone :)

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Device.scala
simao: we'd have to add the uuid to DeviceT and possibly create a new object transport object for the create endpoint, so lets not worry about that now and leave it for a next refactor that includes FE.

PR #55, opened 2018-10-24
Parse and index installation report events.

houcros:


src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Codecs.scala
```diff
@@ -1,13 +1,19 @@
...
 object Codecs {
   private[this] implicit val deviceIdEncoder = Encoder.encodeString.contramap[Device.DeviceOemId](_.underlying)
   private[this] implicit val deviceIdDecoder = Decoder.decodeString.map(Device.DeviceOemId.apply)
 
   implicit val deviceTEncoder = io.circe.generic.semiauto.deriveEncoder[DeviceT]
   implicit val deviceTDecoder = io.circe.generic.semiauto.deriveDecoder[DeviceT]
+
+  implicit val ecuSerialDecoder: Decoder[EcuSerial] =
+    Decoder.decodeString.map(refineV[ValidEcuSerial](_).valueOr(s => throw new IllegalArgumentException(s"Failed to decode EcuSerial $s")))
 }
```
simao: There is a decoder for refined in `com.advancedtelematic.libats.codecs.CirceRefined._`

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
```diff
@@ -3,30 +3,57 @@ package com.advancedtelematic.ota.deviceregistry.db
...
+    IndexedEvent(event.deviceUuid, event.eventId, IndexedEventType.DownloadComplete, None, None, None).asRight
+
+  private def parseInstallationComplete(event: Event): EventIndexResult =
+    event.payload.hcursor.get[CorrelationId]("correlationId")
+      .map(cid => IndexedEvent(event.deviceUuid, event.eventId, IndexedEventType.withName(event.eventType.id), cid.some, None, None))
+      .leftMap(toErrorString(event))
+
+  private def parseInstallationReport(event: Event): EventIndexResult = {
+    val c = event.payload.hcursor
+    c.get[CorrelationId]("correlationId").flatMap { cid =>
```
simao: usually when there are nested flatMaps/maps there is a cleaner for comprehension:

```scala
   for {
      cid <- cursor.get[CorrelationId]("correlationId")
      rc <- cursor.get[ResponseCode]("responseCode")
    } yield IndexedEvent(event.deviceUuid, event.eventId, IndexedEventType.InstallationReport, cid.some, None, rc.some)
```

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
```diff
@@ -3,30 +3,57 @@ package com.advancedtelematic.ota.deviceregistry.db
...
       parseDownloadComplete(event)
     case EventType("InstallationComplete", 0) =>
       parseInstallationComplete(event)
+    case EventType("InstallationReport", 0) =>
+      parseInstallationReport(event)
+    case EventType("EcuInstallationReport", 0) =>
+      parseEcuInstallationReport(event)
     case eventType =>
       s"Unknown event type $eventType".asLeft
   }
```
simao: I wonder if it would be better to delegate the partial function the the parsing functions, would maybe give a cleaner solution. Something like this: https://github.com/advancedtelematic/ota-device-registry/commit/b834c031783883e2a0a721f0b167b0f586e4393f

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Codecs.scala
houcros: Thanks, I figured we had something like that in libats but I couldn't find it :)

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
```diff
@@ -15,45 +15,46 @@ object EventIndex {
...
-    c.get[CorrelationId]("correlationId").flatMap { cid =>
-        c.get[ResponseCode]("responseCode").flatMap { rc =>
-          c.get[EcuSerial]("ecuSerial").map { ecu =>
-            IndexedEvent(event.deviceUuid, event.eventId, IndexedEventType.EcuInstallationReport, cid.some, ecu.some, rc.some)
-          }
-        }
-    }.leftMap(toErrorString(event))
+  private def parseEcuInstallationReport: PartialFunction[Event, EventIndexResult] = {
+    case event if event.eventType == EventType("EcuInstallationReport", 0) =>
+      val c = event.payload.hcursor
```
simao: `val cursor` should be more readable

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
```diff
@@ -15,45 +15,46 @@ object EventIndex {
...
+      val ie = for {
+        cid <- c.get[CorrelationId]("correlationId")
+        rc  <- c.get[ResponseCode]("responseCode")
+        ecu <- c.get[EcuSerial]("ecuSerial")
+      } yield IndexedEvent(event.deviceUuid, event.eventId, IndexedEventType.EcuInstallationReport, cid.some, ecu.some, rc.some)
+      ie.leftMap(toErrorString(event))
```
simao: also maybe `ie` should have a better name.

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
houcros: That was a nice idea, thanks :)

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
houcros: Done.

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
houcros: Done.

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/EventIndex.scala
houcros: Indeed :) Thanks

houcros:
Sure, will squash we've decided we're good to go.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
```diff
@@ -1,19 +1,28 @@
...
 
-    val DownloadComplete, InstallationComplete = Value
+    val DownloadComplete, InstallationComplete, InstallationReport, EcuInstallationReport = Value
+  }
+
+  object EventField extends Enumeration {
+    type EventField
+    val CORRELATION_ID = Value("correlationId")
+    val ECU_SERIAL     = Value("ecu")
+    val RESULT_CODE    = Value("result_code")
```
jerrytrieu: Let’s use ‘resultCode’

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -229,9 +229,8 @@ class DevicesResource(
               }
             }
           } ~
-          (get & parameter('correlationId.as[CorrelationId].?) & pathEnd) { correlationId =>
-            val events = eventJournal.getEvents(uuid, correlationId)
-            complete(events)
+          (get & parameters(('correlationId.as[CorrelationId].?, 'ecuSerial.as[EcuSerial].?, 'resultCode.as[ResultCode].?)) & pathEnd) {
```
jerrytrieu: Need to search all devices as well.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
```diff
@@ -1,19 +1,28 @@
...
 import com.advancedtelematic.ota.deviceregistry.data.DataType.IndexedEventType.IndexedEventType
 import com.advancedtelematic.ota.deviceregistry.data.Device.{DeviceName, DeviceOemId, DeviceType}
 
 object DataType {
-  case class IndexedEvent(device: DeviceId, eventID: String, eventType: IndexedEventType, correlationId: Option[CorrelationId])
+  case class IndexedEvent(device: DeviceId, eventID: String, eventType: IndexedEventType, correlationId: Option[CorrelationId],
+                          ecuSerial: Option[EcuSerial], resultCode: Option[ResultCode])
 
   case class CorrelationId(id: String) extends AnyVal
+  case class ResultCode(value: String) extends AnyVal
```
jerrytrieu: Let’s keep this as Int.

houcros: Removed `ResultCode` and using `Int` instead.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
houcros: Updated.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
houcros: WDYT if we create a new endpoint to search by this params for all devices? Else we have to add this params to the `searchDevice` method, which is getting already a bit complex.

simao: All devices from correlationid ????

houcros: Added new endpoint `/api/v1/devices/events?correlationId=<...>&ecu=<...>&resultCode=<...>`. Is that what we need?

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
houcros: Also: not sure what the range is for `ResultCode`, but I suppose `SMALLINT` should be enough?

PR #58, opened 2018-11-02
Add support to fetch ungrouped devices by group type.

houcros:


src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -92,8 +92,8 @@ class DevicesResource(
         complete(db.run(DeviceRepository.findByDeviceId(ns, oemId)))
       case (None, Some(true), gt, None, rx, o, l) =>
         complete(db.run(DeviceRepository.searchGrouped(ns, gt, rx, o, l)))
-      case (None, Some(false), _, None, rx, o, l) =>
-        complete(db.run(DeviceRepository.searchUngrouped(ns, rx, o, l)))
+      case (None, Some(false), gt, None, rx, o, l) =>
+        complete(db.run(DeviceRepository.searchUngrouped(ns, gt, rx, o, l)))
       case (None, _, _, gid, rx, o, l) =>
         complete(db.run(DeviceRepository.search(ns, rx, gid, o, l)))
       case (Some(_), _, _, Some(_), None, _, _) =>
```
simao: I think these `parameters` reached its limits for complexity..

I think we should use case class extraction to handle this better, something like:

```scala
  case class SearchParams(deviceOemId: DeviceOemId, groupd: Boolean, groupType: GroupType, groupId: GroupId, regex: Refined[String, Regex], offeset: Long, limit: Long)

  def searchDevice(ns: Namespace): Route =
    parameters((
      'deviceId.as[DeviceOemId].?,
      'grouped.as[Boolean].?,
      'groupType.as[GroupType].?,
      'groupId.as[GroupId].?,
      'regex.as[String Refined Regex].?,
      'offset.as[Long].?,
      'limit.as[Long].?)).as(SearchParams)
    { params =>
...
} 
```

This would also allow for validations on SearchParams, see https://doc.akka.io/docs/akka-http/current/routing-dsl/case-class-extraction.html 

Then the SearchParams instance can just be passed to `DeviceRepository` which itself builds the proper query and not the resource.

houcros: Thanks for the tip! That's quite nicer indeed, this params where getting out of control :)

houcros:
As discussed, this would break the device-gateway, so I'm opening a PR there as well.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
```diff
@@ -25,4 +29,9 @@ object DataType {
   implicit val eventShow: Show[Event] = Show { event =>
     s"(device=${event.deviceUuid},eventId=${event.eventId},eventType=${event.eventType})"
   }
+
+  final case class SearchParams(oemId: Option[DeviceOemId], grouped: Option[Boolean], groupType: Option[GroupType],
+                          groupId: Option[GroupId], regex: Option[String Refined Regex], offset: Option[Long], limit: Option[Long]) {
+    require(oemId.isEmpty || (groupId.isEmpty && regex.isEmpty), "Invalid parameter combination.")
```
simao: would be better to make this separate `require`s so we can get different error messages

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/DeviceRepository.scala
```diff
@@ -124,39 +124,41 @@ object DeviceRepository {
...
+        runQueryWithRegex(ns, groupedDevicesQuery(gt), rx)
+
+      case SearchParams(None, Some(false), gt, None, rx, _, _) =>
+        val ungroupedDevicesQuery = devices.filterNot(_.uuid.in(groupedDevicesQuery(gt).map(_.uuid)))
+        runQueryWithRegex(ns, ungroupedDevicesQuery, rx)
+
+      case SearchParams(None, _, _, gid, rx, _, _) =>
+        searchQuery(ns, rx, gid)
+
+      case _ => ???
```
simao: should then throw a `IllegalArgumentsException`

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
```diff
@@ -25,4 +29,10 @@ object DataType {
   implicit val eventShow: Show[Event] = Show { event =>
     s"(device=${event.deviceUuid},eventId=${event.eventId},eventType=${event.eventType})"
   }
+
+  final case class SearchParams(oemId: Option[DeviceOemId], grouped: Option[Boolean], groupType: Option[GroupType],
+                          groupId: Option[GroupId], regex: Option[String Refined Regex], offset: Option[Long], limit: Option[Long]) {
+    require(oemId.isEmpty || groupId.isEmpty, "Invalid parameter combination: deviceId, groupId.")
+    require(oemId.isEmpty || regex.isEmpty, "Invalid parameter combination: deviceId, regex.")
+  }
```
simao: hm. If oemId.isEmpty is false, then the second require only depends on `regex.isEmpty`...

So this would mean that if `oemId.isEmpty` cannot ever be true, so it should have it's own require

This also means that this would be equivalent to `oemId.isEmpty || groupId.isEmpty || regex.isEmpty`, not sure that's what you want ?

houcros: Right! That's not what we want. Then I think the simpler way to put it is:
```
if (oemId.isDefined) {
      require(groupId.isEmpty, "Invalid parameter combination: deviceId, groupId.")
      require(regex.isEmpty, "Invalid parameter combination: deviceId, regex.")
    }
```

simao:
```scala
      require(oemId.isDefined && groupId.isEmpty, "GroupId cannot be empty when oemId is defined")
      require(oemId.isDefined && regex.isEmpty, "Regex cannot be empty when oemId is defined")
```

But this should probably have been caught by tests

houcros: Updated error messages.

PR #60, opened 2018-11-13
Calculate update report counts for devices and ecus per result code

houcros:


houcros:
(Will squash the commits eventually.)

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -180,6 +188,9 @@ class DevicesResource(
           case None      => complete(Errors.InvalidGroupExpression(""))
           case Some(exp) => countDynamicGroupCandidates(ns.namespace, exp)
         } ~
+        (path("stats") & parameters('correlationId.as[CorrelationId], 'level.as[String].?)) { (cid, filterLevel) =>
```
houcros: Not too happy about this `level` parameter. Any idea to better distinguish the requests when we want the count per devices or per ECUs?

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
```diff
@@ -12,8 +12,14 @@ import eu.timepit.refined.string.Regex
 object DataType {
   case class IndexedEvent(device: DeviceId, eventID: String, eventType: IndexedEventType, correlationId: Option[CorrelationId])
 
-  case class CorrelationId(id: String) extends AnyVal
+  case class UpdateStat(resultCode: Int, total: Int)
 
+  case class CorrelationId(id: String) extends AnyVal
+  object CorrelationId {
```
houcros: I took this from director. Should I maybe move it to libats for common use?

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/UpdateReportRepository.scala
```diff
@@ -0,0 +1,71 @@
...
+      for {
+        _ <- deviceUpdateReports += DeviceReport(deviceUuid, correlationId, deviceResultCode)
+        _ <- ecuUpdateReports ++= operationResults.map {
+          case (ecuSerial, operationResult) => EcuReport(ecuSerial, correlationId, operationResult.resultCode)
+        }
+      } yield ()
+    q.transactionally
+  }
+
+  def calculateFailedDevicesStats(correlationId: CorrelationId)(implicit ec: ExecutionContext): DBIO[Seq[UpdateStat]] = {
```
houcros: @simao this two queries are identical but on different tables. Do you know if there is a slick way to do it and avoid this ugly code duplication?

simao: This can be abstracted like this: https://github.com/advancedtelematic/ota-device-registry/commit/7f0dbecae874f7d6d2ec49fd7cd9dd3784576668

But, if you are are trying to do the same query on two different tables maybe that means everything should be on the same table?

build.sbt
```diff
@@ -50,6 +52,7 @@ lazy val library =
       "libats-metrics-akka",
       "libats-metrics-prometheus"
     ).map("com.advancedtelematic" %% _ % Version.libAts)
+    val libTuf = "com.advancedtelematic" %% "libtuf-server" % Version.libTuf
```
simao: Why do we need this here? This is kind of a big dependency we are pulling..

src/main/scala/com/advancedtelematic/ota/deviceregistry/daemon/DeviceUpdateReportListener.scala
```diff
@@ -0,0 +1,34 @@
...
+import com.advancedtelematic.ota.deviceregistry.db.UpdateReportRepository
+import com.codahale.metrics.MetricRegistry
+import com.typesafe.config.Config
+import slick.jdbc.MySQLProfile.api._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class DeviceUpdateReportListener(config: Config, db: Database, metrics: MetricRegistry)
+                                (implicit val ec: ExecutionContext, system: ActorSystem) {
+  def start(): Unit = {
```
simao: I don't know why this is like this in device registry, but we should abstract this start boiler plate like other services do, check out for example https://github.com/advancedtelematic/campaigner/blob/master/src/main/scala/com/advancedtelematic/campaigner/DaemonBoot.scala#L47

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
simao: yes

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/UpdateReportRepository.scala
houcros: Yes, I did consider and tried putting everything in the same table, where either `deviceUuid` or `ecuSerial` would be defined but not both. Looked cleaner, but I talked to Jerry and probably the tables will start diverging soon, so seems a tricky situation...

build.sbt
houcros: To catch the `com.advancedtelematic.libtuf_server.data.Messages.DeviceUpdateReport` messages.

simao: ok

simao: so those messages should be in libats and not libtuf....

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/UpdateReportRepository.scala
simao: ok

build.sbt
houcros: But `DeviceUpdateReport` uses `OperationResult` in one of its members, and the `OperationResult` from libats is deprecated in favour of the one in libtuf. So then we'd have to include libtuf as a dependency of libats, which would be uglier 😕

simao: Then maybe we need to undeprecate OperationResult in libats ;)

houcros: We'd also need libtuf if we want to use `MessageListenerSupport#startListener` and reduce the boilerplate in the listener as you suggest.

simao: that should definitely not be in libtuf...

houcros: Sounds like we need a big refactor... 😅

houcros: Sorry, that's in libats, my bad.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
simao: you could use an enum?  `level=device|ecu`

houcros: Can do that :) Although my concern was that maybe it didn't look too "RESTy" to filter the ECU counts like this. But if it's good enough for now I'll make it an enum and move on :)

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE DeviceReport (
```
jerrytrieu: Let's call it `DeviceInstallationReport`

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
```diff
@@ -0,0 +1,11 @@
...
+    device_uuid char(36) COLLATE utf8_bin NOT NULL,
+    correlation_id varchar(256) NOT NULL,
+    result_code SMALLINT NOT NULL
+);
+
+CREATE TABLE EcuReport (
```
jerrytrieu: `EcuInstallationReport`

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
```diff
@@ -0,0 +1,11 @@
...
+    ecu_serial VARCHAR(64)  NOT NULL,
```
jerrytrieu: Let's make `PRIMARY KEY (device_uuid, ecu_id)` for future refactoring.

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE DeviceReport (
+    device_uuid char(36) COLLATE utf8_bin NOT NULL,
+    correlation_id varchar(256) NOT NULL,
+    result_code SMALLINT NOT NULL
```
jerrytrieu: Let's have `Json` blob `installation_report` everything from the message.

koshelev: Or probably id of an `DeviceEvent` containing the blob?

koshelev: Why do we have `result_code` numeric? Why not a string?

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/UpdateReportRepository.scala
houcros: Thanks for the lesson :) Abstracted like that.

build.sbt
houcros: Removed libtuf as dependency.

src/main/scala/com/advancedtelematic/ota/deviceregistry/daemon/DeviceUpdateReportListener.scala
koshelev: Or even kill this daemon concept altogether...

simao: There is no daemon in this service... But I am not sure no daemon is a good idea... This stuff should not happen on the same app that serves http to client

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
jerrytrieu: We need to support this for current aktualizr. There's also `result_text` in the message, used as description.

koshelev: we always can convert a number to a string :)
The problem is, some customers have result codes as strings.

src/main/scala/com/advancedtelematic/ota/deviceregistry/daemon/DeviceUpdateReportListener.scala
koshelev: It depends. If both daemon and service are modifying same aggregates then there is not reason to separate them unless message processing is really resource-consuming.

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
houcros: Updated.

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
houcros: Updated.

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
koshelev: @houcros we should not rely in ecu having an globally unique serial number. We don't really care of ecu being globally identifiable. What we need is to be able to identify ecu inside a device(a vehicle). This identifier is not always a serial number, that's why `ecu_id`. And as it is not globally unique we will need `device_uuid`

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
houcros: Using `CorrelationId` from libats. Probably will have to update a bit after Jerry merges his PR in libats.

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
houcros: Oh I see now. Will update this.

src/main/scala/com/advancedtelematic/ota/deviceregistry/daemon/DeviceUpdateReportListener.scala
koshelev: Sure. Why I would like to avoid daemons: it is an additional service to take care and the fact that two services share the same schema and one of them is responsible for migrations makes me feel bit uncomfortable.

simao: Migrations **need** to run async, at the least the way we run them. If migration takes a bit longer the health check will fail and k8s will kill the container, interrupting the migration.

koshelev: yes, I know. what i mean is the case of services with different versions relying on different db schema get deployed. But let's discuss it offline :)

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
houcros: Still, do we want to include the `correlationId` in the PK? I understand we're not supposed to have more than one report per device/ECU.

koshelev: We can, but probably just `FOREIGN KEY (device_uuid, correlation_id)` would be enough.

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
```diff
@@ -0,0 +1,17 @@
...
+    device_uuid char(36) COLLATE utf8_bin NOT NULL,
+
+    PRIMARY KEY (correlation_id, device_uuid)
+);
+
+CREATE TABLE EcuInstallationReport (
+    correlation_id varchar(256) NOT NULL,
+    result_code SMALLINT NOT NULL,
+    device_uuid char(36) COLLATE utf8_bin NOT NULL,
+    ecu_serial VARCHAR(64)  NOT NULL,
```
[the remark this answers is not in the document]
houcros: Sure, done.

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
```diff
@@ -0,0 +1,17 @@
...
+
+    PRIMARY KEY (device_uuid, ecu_serial),
```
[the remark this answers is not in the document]
houcros: I added `correlation_id`.

src/main/scala/com/advancedtelematic/ota/deviceregistry/daemon/DeviceUpdateReportListener.scala
houcros: I refactored to use `startListener` like we do in campaigner (still in app). I'll try and open a refactor PR to do the same with the remaining listeners.

src/main/resources/db/migration/V27__device_and_ecu_installation_reports.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE DeviceInstallationResult (
+    correlation_id varchar(256) NOT NULL,
+    result_code VARCHAR(256) NOT NULL,
+    device_uuid char(36) COLLATE utf8_bin NOT NULL,
+    installation_report JSON,
+
```
jerrytrieu: Let’s have ‘recevedAt’ column

src/main/resources/db/migration/V28__device_and_ecu_installation_reports.sql
houcros: `resultCode` is now a String.

src/main/resources/db/migration/V27__device_and_ecu_installation_reports.sql
houcros: Added.

PR #61, opened 2018-11-19
Refactor listeners and DeviceResource route.

houcros:
- Refactor listeners to reduce boilerplate.
- Refactor DeviceResource extracting common GET method in routes.

### DIFF
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
@@ -214,26 +214,26 @@ class DevicesResource(
         }
       } ~
       deviceNamespaceAuthorizer { uuid =>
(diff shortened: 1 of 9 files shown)


src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -214,26 +214,38 @@ class DevicesResource(
...
-        } ~
-        (scope.delete & pathEnd) {
-          deleteDevice(ns.namespace, uuid)
-        } ~
-        (scope.get & pathEnd) {
-          fetchDevice(uuid)
-        } ~
-        (scope.get & path("groups") & pathEnd) {
+      scope.get {
+        (path("groups") & pathEnd) {
```
simao: This should work without `pathEnd`, same below

PR #63, opened 2018-11-26
Migrate old installation reports from auditor to device-registry.

houcros:


src/main/scala/com/advancedtelematic/ota/deviceregistry/db/MigrateOldInstallationReports.scala
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  def run: Future[Done] = {
+    val source = db.stream(devices.map(d => (d.namespace, d.uuid)).result)
+    Source
+      .fromPublisher(source)
+      .mapAsyncUnordered(1) { case (ns, did) => auditor.getUpdateReports(ns, did)}
+      .map(_.map(toNewSchema))
+      .map(_.map(saveReport))
+      .map(DBIO.sequence(_))
```
simao: should probably be a transaction for each device ?

houcros: Updated.

houcros:
Right! I extended the last test in the migration to check that it actually is idempotent.

PR #64, opened 2018-11-30
Move ECU info to dev-reg and add an endpoint to fetch it.

houcros:
Sorry for the huge PR! We're now storing the ECU info in device-registry, so in this PR:

- Now we require the ECU info when creating a new device.
- Migrate the ECU info from director to device-registry.
- Create a new endpoint to get the ECU info.

**Note: This changes the way we create devices, so other services might need to change when we merge this?**

### DIFF
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
@@ -146,6 +146,11 @@ class DevicesResource(
   def countDynamicGroupCandidates(ns: Namespace, expression: GroupExpression): Route =
     complete(db.run(DeviceRepository.countDevicesForExpression(ns, expression)))
 
+  def getEcusForDevice(ns: Namespace, deviceId: DeviceId): Route =
+    parameters(('offset.as[Long].?, 'limit.as[Long].?)) { (offset, limit) =>
+      complete(db.run(EcuRepository.listEcusForDevice(deviceId, offset, limit)))
+    }
+
   def getGroupsForDevice(ns: Namespace, uuid: DeviceId): Route =
     parameters(('offset.as[Long].?, 'limit.as[Long].?)) { (offset, limit) =>
       complete(db.run(GroupMemberRepository.listGroupsForDevice(ns, uuid, offset, limit)))
@@ -238,6 +243,9 @@ class DevicesResource(
       } ~
       deviceNamespaceAuthorizer { uuid =>
         scope.get {
+          path("ecus") {
+            getEcusForDevice(ns.namespace, uuid)
+          } ~
           path("groups") {
             getGroupsForDevice(ns.namespace, uuid)
           } ~
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
@@ -31,7 +31,7 @@ import com.advancedtelematic.ota.deviceregistry.DevicesResource.EventPayload
 import com.advancedtelematic.ota.deviceregistry.common.Errors
 import com.advancedtelematic.ota.deviceregistry.data.Codecs._
 import com.advancedtelematic.ota.deviceregistry.data.DataType.InstallationStatsLevel.InstallationStatsLevel
-import com.advancedtelematic.ota.deviceregistry.data.DataType.{DeviceT, InstallationStatsLevel, SearchParams, UpdateDevice}
+import com.advancedtelematic.ota.deviceregistry.data.DataType.{CreateDevice, InstallationStatsLevel, SearchParams, UpdateDevice}
 import com.advancedtelematic.ota.deviceregistry.data.Device.{ActiveDeviceCount, DeviceOemId}
 import com.advancedtelematic.ota.deviceregistry.data.Group.{GroupExpression, GroupId}
 import com.advancedtelematic.ota.deviceregistry.data.GroupType.GroupType
@@ -113,22 +113,24 @@ class DevicesResource(
       'limit.as[Long].?)).as(SearchParams)
     { params => complete(db.run(DeviceRepository.search(ns, params))) }
 
-  def createDevice(ns: Namespace, device: DeviceT): Route = {
-    val f = db
-      .run(DeviceRepository.create(ns, device))
-      .andThen {
-        case scala.util.Success(uuid) =>
-          messageBus.publish(
-            DeviceCreated(ns, uuid, device.deviceName, device.deviceId, device.deviceType, Instant.now())
-          )
-      }
+  def createDevice(ns: Namespace, payload: CreateDevice): Route =
+    if (payload.ecus.exists(_.ecuId == payload.primaryEcu)) {
+      val f = db
+        .run(DeviceRepository.create(ns, payload))
+        .flatMap { uuid =>
+          messageBus
+            .publishSafe(DeviceCreated(ns, uuid, payload.deviceName, payload.deviceId, payload.deviceType, Instant.now()))
+            .map(_ => uuid)
+        }
 
-    onSuccess(f) { uuid =>
-      respondWithHeaders(List(Location(Uri("/devices/" + uuid.show)))) {
-        complete(Created -> uuid)
+      onSuccess(f) { uuid =>
+        respondWithHeaders(List(Location(Uri("/devices/" + uuid.show)))) {
+          complete(Created -> uuid)
+        }
       }
+    } else {
+      failWith(Errors.MissingPrimaryEcu)
     }
-  }
 
   def deleteDevice(ns: Namespace, uuid: DeviceId): Route = {
     val f = messageBus.publish(DeleteDeviceRequest(ns, uuid, Instant.now()))
@@ -216,8 +218,8 @@ class DevicesResource(
   def api: Route = namespaceExtractor { ns =>
     val scope = Scopes.devices(ns)
     pathPrefix("devices") {
-      (scope.post & entity(as[DeviceT]) & pathEnd) { device =>
-        createDevice(ns.namespace, device)
+      (scope.post & entity(as[CreateDevice])) { device =>
+          createDevice(ns.namespace, device)
       } ~
       scope.get {
         (path("count") & parameter('expression.as[GroupExpression].?)) {
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/client/DirectorClient.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/client/DirectorClient.scala
@@ -1,0 +1,51 @@
+package com.advancedtelematic.ota.deviceregistry.client
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.model._
+import akka.stream.Materializer
+import cats.syntax.show._
+import com.advancedtelematic.libats.codecs.CirceCodecs.{refinedDecoder, refinedEncoder}
+import com.advancedtelematic.libats.data.DataType.Namespace
+import com.advancedtelematic.libats.data.EcuIdentifier
+import com.advancedtelematic.libats.http.Errors.RemoteServiceError
+import com.advancedtelematic.libats.http.HttpOps.HttpRequestOps
+import com.advancedtelematic.libats.http.ServiceHttpClient
+import com.advancedtelematic.libats.messaging_datatype.DataType.DeviceId
+import com.advancedtelematic.libtuf.data.TufCodecs._
+import com.advancedtelematic.libtuf.data.TufDataType.{HardwareIdentifier, TufKey}
+import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport._
+import io.circe.Decoder
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.{ExecutionContext, Future}
+
+final case class EcuInfo(id: EcuIdentifier, hardwareId: HardwareIdentifier, primary: Boolean)
+
+trait DirectorClient {
+  def getEcuInfo(ns: Namespace, deviceId: DeviceId): Future[Seq[EcuInfo]]
+  def getPublicKey(ns: Namespace, deviceId: DeviceId, ecuIds: EcuIdentifier): Future[TufKey]
+}
+
+class DirectorHttpClient(uri: Uri, httpClient: HttpRequest => Future[HttpResponse])
+                       (implicit ec: ExecutionContext, system: ActorSystem, mat: Materializer) extends ServiceHttpClient(httpClient) with DirectorClient {
+
+  private val _log = LoggerFactory.getLogger(this.getClass)
+  private val pathPrefix = uri.path / "api" / "v1" / "admin" / "devices"
+  implicit val ecuDecoder: Decoder[EcuInfo] = io.circe.generic.semiauto.deriveDecoder
+
+  override def getEcuInfo(ns: Namespace, deviceId: DeviceId): Future[Seq[EcuInfo]] = {
+    val path = pathPrefix / deviceId.show
+    val req = HttpRequest(HttpMethods.GET, uri.withPath(path)).withNs(ns)
+    execHttp[Seq[EcuInfo]](req)().recover {
+      case e: RemoteServiceError =>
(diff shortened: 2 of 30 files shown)


src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Ecu.scala
```diff
@@ -0,0 +1,22 @@
...
+import java.security.PublicKey
+
+import com.advancedtelematic.libats.codecs.CirceCodecs.{refinedDecoder, refinedEncoder}
+import com.advancedtelematic.libats.data.DataType.Namespace
+import com.advancedtelematic.libats.messaging_datatype.DataType.{DeviceId, EcuSerial}
+import com.advancedtelematic.libtuf.data.TufCodecs._
+import com.advancedtelematic.libtuf.data.TufDataType.{HardwareIdentifier, KeyType, TufKey}
+import io.circe.{Decoder, Encoder}
+
+final case class Ecu(namespace: Namespace, deviceUuid: DeviceId, ecuId: EcuSerial,
```
koshelev: Can we probably fix the types and have `ecuId: EcuIdentifier` and `ecuType: EcuType`?

src/main/resources/db/migration/V28__create_ecu_table.sql
```diff
@@ -0,0 +1,12 @@
...
+  namespace   VARCHAR(200)  NOT NULL,
+  device_uuid CHAR(36)      NOT NULL COLLATE utf8_bin,
+  ecu_id      VARCHAR(64)   NOT NULL,
+  ecu_type    VARCHAR(200)  NULL,
+  `primary`   BOOL          NOT NULL,
+  key_type    VARCHAR(16)   NOT NULL,
+  public_key  VARCHAR(4096) NOT NULL,
+
+  PRIMARY KEY (device_uuid, ecu_id),
+  CONSTRAINT `fk_ecu_device` FOREIGN KEY (device_uuid) REFERENCES Device(uuid)
```
koshelev: `ON DELETE CASCADE`?

src/main/resources/db/migration/V28__create_ecu_table.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE Ecu (
+  namespace   VARCHAR(200)  NOT NULL,
```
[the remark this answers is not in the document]
houcros: Yes, it seems we can do without it. I'll remove it.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Ecu.scala
houcros: You're proposing to change the types in `libats` and `libtuf`, right? Sure, can do that.

src/main/resources/db/migration/V28__create_ecu_table.sql
houcros: I figured it was not necessary because I added this line: https://github.com/advancedtelematic/ota-device-registry/pull/64/files#diff-9d4bb5efdc877024f0457b71834462ebR218
Am I mistaken?

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Ecu.scala
koshelev: Yes.
We probably still can have `type EcuSerial = EcuIdentifier` for backwards compatibility. But as we are on it anyway, probably we also should get rid of `Refined`. @simao, WDYT?

src/main/resources/db/migration/V28__create_ecu_table.sql
koshelev: No, you are right, should be enough

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/Ecu.scala
houcros: > probably we also should get rid of Refined.
What do you mean exactly? We don't want validation on `EcuType` and `EcuIdentifier` anymore? Or only one of them?

koshelev: We do need validation, but refined is probably not the best way to do it. We can have something like
```
final case class EcuIdentifier(value: String) extends AnyVal
object EcuIdentifier {
  def fromString(value: String): Either[ValidationError, EcuIdentifier] = ????
}
```

simao: I also don't like `Refined`, but I think the reason we keep using it is because it automagically gives us Validations, circe Codecs, akka-http path matchers and slick mappings, which simplifies our work a lot.

I think to not use `Refined` we will need to offer this functionality for other types. IIRC, there is already support for AnyVal with circe codecs and slick mappings, maybe we could extend that to support validations and akka-http path matchers. I think that would be better than to implement ad-hoc factory/validation methods.

koshelev: Validations need to be implemented explicitly (we do it for refined anyway, right?). I can look into how we can derive the rest from it

houcros: I'm not sure that we have path matchers for `Refined`. At least I didn't find any in `libats`. So it seems that actually our support for `AnyVal` might be as almost good as the one for `Refined`.
I've opened https://github.com/advancedtelematic/libats/pull/82 to address this, WDYT?

simao: For refined we implement a codec that uses `refineV` which validates the refined.

simao: @houcros yes, lets do that :+1:

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -104,9 +105,10 @@ class DevicesResource(
       'limit.as[Long].?)).as(SearchParams)
     { params => complete(db.run(DeviceRepository.search(ns, params))) }
 
-  def createDevice(ns: Namespace, device: DeviceT): Route = {
+  @deprecated("Remove when dropping support for the old way of creating devices.")
```
simao: Why don't we just remove this now and get ride of `DeviceT` ?

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -121,6 +123,26 @@ class DevicesResource(
     }
   }
 
+  def createDevice(ns: Namespace, payload: CreateDevice): Route = {
+
+    if (!payload.ecus.map(_.ecuId).contains(payload.primaryEcu)) complete(Errors.MissingPrimaryEcu)
```
simao: `failWith` instead of `complete`.

simao: And complete should be on next line

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -121,6 +123,26 @@ class DevicesResource(
...
+            DeviceCreated(ns, uuid, payload.deviceName, payload.deviceId, payload.deviceType, Instant.now())
+          )
+      }
+
+    onSuccess(f) { uuid =>
+      respondWithHeaders(List(Location(Uri("/devices/" + uuid.show)))) {
+        complete(Created -> uuid)
+      }
+    }
+  }
```
simao:
```scala
  def createDevice(ns: Namespace, payload: CreateDevice): Route = {
    if (!payload.ecus.map(_.ecuId).contains(payload.primaryEcu))
      failWith(Errors.MissingPrimaryEcu)

    val f = db
      .run(DeviceRepository.create(ns, payload))
      .flatMap { uuid =>
        messageBus.publishSafe(
          DeviceCreated(ns, uuid, payload.deviceName, payload.deviceId, payload.deviceType, Instant.now())
        ).map(_ => uuid)
      }
    
    onSuccess(f) { uuid =>
      respondWithHeaders(List(Location(Uri("/devices/" + uuid.show)))) {
        complete(Created -> uuid)
      }
    }
  }
```

src/main/scala/com/advancedtelematic/ota/deviceregistry/client/DirectorClient.scala
```diff
@@ -0,0 +1,50 @@
+package com.advancedtelematic.ota.deviceregistry.client
```
simao: DirectorClient should be in director where it can be unit tested against director resources.. director needs to be published to nexus and then device-registry can use it. Unfortunately it will pull a lot of dependencies but looks like device registry is already depending on lots of tuf stuff anyway.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
```diff
@@ -31,6 +32,7 @@ object DataType {
     case object Ecu extends InstallationStatsLevel
   }
 
+  @deprecated("Remove when dropping support for the old way of creating devices.")
   final case class DeviceT(deviceName: DeviceName,
```
simao: Lets remove it now?

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/MigrateEcuInfoFromDirector.scala
```diff
@@ -0,0 +1,42 @@
...
+        Future.traverse(s) {
+          case (ns, did, ecuInfo) =>
+            director
+              .getPublicKey(ns, did, ecuInfo.id)
+              .map(Ecu(did, ecuInfo.id, ecuInfo.hardwareId, ecuInfo.primary, _))
+        }
+      }
+      .runForeach(_.foreach(saveEcu))
+  }
+}
```
simao:
```scala
  def saveEcu(ecu: Ecu): Future[EcuSerial] = {
    _log.info(s"Migrating ECU from director: $ecu.")
    db.run(EcuRepository.ecus.insertOrUpdate(ecu).map(_ => ecu.ecuId))
  }

  def run: Future[Done] = {
    val source = db.stream(devices.map(d => (d.namespace, d.uuid)).result)
    Source
      .fromPublisher(source)
      .mapAsync(1) { case (ns, did) => director.getEcuInfo(ns, did).map(_.map((ns, did, _))) }
      .mapAsync(1) { s =>
        Future.traverse(s) {
          case (ns, did, ecuInfo) =>
            for {
              pubKey <- director.getPublicKey(ns, did, ecuInfo.id)
              ecuId <- saveEcu(Ecu(did, ecuInfo.id, ecuInfo.hardwareId, ecuInfo.primary, pubKey))
            } yield ecuId
        }
      }
      .runWith {
        Sink.foreach { ecuId =>
          _log.info(s"Migrated ecu $ecuId")
        }
      }
  }
```

src/test/scala/com/advancedtelematic/ota/deviceregistry/util/FakeDirectorClient.scala
```diff
@@ -0,0 +1,34 @@
...
+class FakeDirectorClient extends DirectorClient {
+
+  private val ecus = new ConcurrentHashMap[(Namespace, DeviceId, EcuSerial), Ecu]().asScala
+
+  def addEcus(ns: Namespace, deviceId: DeviceId, newEcus: Ecu*): Unit =
+    newEcus.foreach(ecu => ecus.put((ns, deviceId, ecu.ecuId), ecu))
+
+  override def getEcuInfo(ns: Namespace, did: DeviceId): Future[Seq[EcuInfo]] = FastFuture.successful {
+    ecus
+      .filter(t => t._1._1 == ns && t._1._2 == did)
```
simao: `.filter{ case ((_ns, deviceId, _), _)  => _ns == ns && deviceId == did }`

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
houcros: There was backward-compatibility concern. But we've talked it over and thankfully yes, let's get rid of this!

houcros: Removed.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
houcros: Updated.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
houcros: Updated.
(I had to use an if/else because, against what I would expect, after `failWith` it would execute the rest of the code and create the device 😕.)

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
houcros: Removed.

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/MigrateEcuInfoFromDirector.scala
houcros: Updated.

src/test/scala/com/advancedtelematic/ota/deviceregistry/util/FakeDirectorClient.scala
houcros: Updated.

src/main/scala/com/advancedtelematic/ota/deviceregistry/client/DirectorClient.scala
houcros: We discussed with @jerrytrieu and it seems it would be too much overhead to move the client. After all, it's a temporary thing. We just need it for this migration and we could get rid of it after all the ECUs have been moved.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
simao: And with complete as well...

houcros: Yeah, my bad, I was assuming that it would complete the request and return from the method.

simao: those methods return a `Route`, that needs to be executed. If you don't return that route anywhere then it's not used.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/DataType.scala
```diff
@@ -53,4 +49,13 @@ object DataType {
       require(regex.isEmpty, "Invalid parameters: regex must be empty when searching by deviceId.")
     }
   }
+
+  final case class CreateEcu(ecuId: EcuSerial, ecuType: HardwareIdentifier, clientKey: TufKey)
+  final case class CreateDevice(deviceName: DeviceName,
+                                deviceId: DeviceOemId,
```
koshelev: Considering we have a type `DeviceId` which we are using for device's uuid, I would prefer to have "externalId` or `providedId` or smth. instead of `deviceId: DeviceOemId`

houcros: What about `manufacturerId`?

koshelev: Naming is hard :) I would interpret `manufacturerId` as serial number or something of device, which is not necessary the case. It is an id owner of a fleet of devices is using to refer to it (or more specifically a string it wrote into device's certificate :) )

houcros:
Needs to be merged after #65 too keep the SQL migrations in sync.

houcros:
1yo and I don't even remember what was the point of this 😄

PR #65, opened 2018-12-11
Save success flag for the installation reports and return in count endpoints.

houcros:


houcros:
For now yes, but AFAIU eventually `success` will come from aktualizr in the installation reports.

jerrytrieu:
Yes `success` will come from aktualizr. Let's merge.

PR #66, opened 2018-12-12
Archive indexed events when deleting a device.

houcros:
There is a FK constraint from the `IndexedEvents` table to the `EventJournal` table, so I was seeing some errors in the logs.

In the future, wouldn't it be an improvement to delete in the SQL when creating the constrains? Or is there any advantage to doing it through Slick?


simao:
The constraint is good for devices that are not deleted, we want to make sure there is a device for each event. Thing is the devices should also not be deleted but only archived, but that is another discussion. I guess the archive table is a shitty thing to have but it's better than deleting the events outright.

houcros:
Alright, I've added a new archive table and store them there before deleting :)

PR #70, opened 2019-01-14
Export installation failures as CSV

houcros:
Minimal export for now: device UUID, device ID and failure code.

### DIFF
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
@@ -11,8 +11,8 @@ package com.advancedtelematic.ota.deviceregistry
 import java.time.{Instant, OffsetDateTime}
 
 import akka.actor.ActorSystem
-import akka.http.scaladsl.model.headers.Location
-import akka.http.scaladsl.model.{StatusCodes, Uri}
+import akka.http.scaladsl.model._
+import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.server._
 import akka.http.scaladsl.unmarshalling.{FromStringUnmarshaller, Unmarshaller}
 import akka.stream.ActorMaterializer
@@ -34,7 +34,7 @@ import com.advancedtelematic.ota.deviceregistry.data.DataType.{DeviceT, Installa
 import com.advancedtelematic.ota.deviceregistry.data.Device.{ActiveDeviceCount, DeviceOemId}
 import com.advancedtelematic.ota.deviceregistry.data.Group.{GroupExpression, GroupId}
 import com.advancedtelematic.ota.deviceregistry.data.GroupType.GroupType
-import com.advancedtelematic.ota.deviceregistry.data.PackageId
+import com.advancedtelematic.ota.deviceregistry.data.{CsvSerializer, PackageId}
 import com.advancedtelematic.ota.deviceregistry.db._
 import com.advancedtelematic.ota.deviceregistry.messages.{DeleteDeviceRequest, DeviceCreated}
 import eu.timepit.refined.api.Refined
@@ -195,6 +195,21 @@ class DevicesResource(
     complete(db.run(action))
   }
 
+  def exportFailureStats(correlationId: CorrelationId): Route = {
+    val csvHeader = CsvSerializer.header("Device UUID", "Device ID", "Failure Code")
+    val f =
+      InstallationReportRepository
+        .fetchDeviceFailures(correlationId)
+        .runFold(csvHeader)((s, t) => s ++ CsvSerializer.asCsvRow(t))
+
+    onSuccess(f) { s =>
+      val h = `Content-Disposition`(ContentDispositionTypes.attachment, Map("filename" -> "device-failures.csv"))
+      respondWithHeader(h) {
+        complete(HttpEntity(ContentTypes.`text/csv(UTF-8)`, s))
+      }
+    }
+  }
+
   def api: Route = namespaceExtractor { ns =>
     val scope = Scopes.devices(ns)
     pathPrefix("devices") {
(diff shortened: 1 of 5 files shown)


src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -206,8 +219,13 @@ class DevicesResource(
           case None      => complete(Errors.InvalidGroupExpression(""))
           case Some(exp) => countDynamicGroupCandidates(ns.namespace, exp)
         } ~
-        (path("stats") & parameters('correlationId.as[CorrelationId], 'reportLevel.as[InstallationStatsLevel].?)) {
-          (cid, reportLevel) => fetchInstallationStats(cid, reportLevel)
+        pathPrefix("stats") {
+          (path("export") & parameter('correlationId.as[CorrelationId])) { cid =>
```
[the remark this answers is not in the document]
houcros: Thanks! Updated to use `Accept` header.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/CsvSerializable.scala
```diff
@@ -0,0 +1,25 @@
...
+  val recordSeparator = "\n"
+
+  def asCsv(header: Seq[String], rows: Seq[CsvSerializable]): String = {
+    val head = header.mkString("", fieldSeparator, recordSeparator)
+    rows
+      .map(_.serialize)
+      .map(_.mkString(fieldSeparator))
+      .mkString(head, recordSeparator, "")
+  }
+}
```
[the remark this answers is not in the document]
houcros: Thanks for pointing this out. I figured there must be a more elegant way.
I've updated but I'm not sure I totally grasped the idea, please see if it makes sense.

houcros:
(Will squash.)

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -206,8 +219,13 @@ class DevicesResource(
           case None      => complete(Errors.InvalidGroupExpression(""))
           case Some(exp) => countDynamicGroupCandidates(ns.namespace, exp)
         } ~
-        (path("stats") & parameters('correlationId.as[CorrelationId], 'reportLevel.as[InstallationStatsLevel].?)) {
-          (cid, reportLevel) => fetchInstallationStats(cid, reportLevel)
+        (path("stats") & parameters('correlationId.as[CorrelationId], 'reportLevel.as[InstallationStatsLevel].?)) { (cid, reportLevel) =>
```
simao: hm. Ok but is not ideal... 

Thing is you should be able to implement an akka Marshaller that marshalls that to CSV and akka-http will deal with the media type negotiation, ie check that the client supports csv or json and if it supports csv uses the csv marshaller otherwise the json marshaller. That's why I said you might need to use `.json` or `.csv` in the url if you are using if from the browser as the browser would probably send something like `*/*` which would fit both csv and json.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/CsvSerializable.scala
```diff
@@ -0,0 +1,25 @@
+package com.advancedtelematic.ota.deviceregistry.data
+
+import cats.syntax.show._
+import com.advancedtelematic.libats.messaging_datatype.DataType.DeviceId
+
+final case class DeviceFailure(deviceId: DeviceId, code: String)
```
simao: Not sure I like this class in this file, it's used from the repo... you could use a tuple if this is not necessary for anything else.

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/CsvSerializable.scala
```diff
@@ -0,0 +1,25 @@
...
+
+trait CsvSerializer[T] {
```
simao: Yeah this was the idea.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
houcros: AFIU this will be called by the FE on a button click, so they should be able to set the header in the request. (But it hasn't really been clarified yet). So I'll try to implement the marshaller and see if I can make this nicer.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -196,9 +196,11 @@ class DevicesResource(
   }
 
   def exportFailureStats(correlationId: CorrelationId): Route = {
-    val f = db
-      .run(InstallationReportRepository.fetchDeviceFailures(correlationId))
-      .map(s => CsvSerializer.asCsv(Seq("Device UUID","Failure Code"), s))
+    val csvHeader = CsvSerializer.header("Device UUID", "Failure Code")
```
kmanikandan-kmk: I think, having DeviceID in the CSV data would be helpful as Device UUID is OTA connect internal identifier for the device which Campaign Manager user does not have any idea about.

houcros: Right! Will try to squeeze that in here.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
houcros: Ah now I see what you mean about the content negotiation being done by the Marshaller. However AFAIU that implies that we're trying to marshal the same kind of data to either JSON or CSV. But in this case the CSV structure is different than (and not a subset of) the JSON. So I suppose that the right thing to do would be to have this in a different endpoint actually?

src/main/scala/com/advancedtelematic/ota/deviceregistry/data/CsvSerializable.scala
houcros: Yes, removed.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
houcros: Done.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -195,6 +195,19 @@ class DevicesResource(
     complete(db.run(action))
   }
 
+  def exportFailureStats(correlationId: CorrelationId): Route = {
```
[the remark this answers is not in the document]
houcros: I've changed this following your tip. Please see if it fits what you had in mind.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
simao: Yes ok then it needs a new endpoint, just some endpoint that describes the resource you are getting rather than how to get it. so could be `/failure-stats` and `installation-stats` for example.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -206,8 +221,15 @@ class DevicesResource(
           case None      => complete(Errors.InvalidGroupExpression(""))
           case Some(exp) => countDynamicGroupCandidates(ns.namespace, exp)
         } ~
-        (path("stats") & parameters('correlationId.as[CorrelationId], 'reportLevel.as[InstallationStatsLevel].?)) {
-          (cid, reportLevel) => fetchInstallationStats(cid, reportLevel)
+        (path("stats") & parameters('correlationId.as[CorrelationId], 'reportLevel.as[InstallationStatsLevel].?)) { (cid, reportLevel) =>
+          optionalHeaderValueByType[Accept]() { accept =>
```
simao: most likely compiler will issue a warning here :\

PR #73, opened 2019-01-29
Hefty refactor

houcros:
I started by moving away from `Refined` to use our new `ValidatedGeneric` but it ended up as a bigger refactor. What this PR mainly does:

- Use `ValidatedGeneric` instead of `Refined` for `DeviceName`.
- Use `ValidatedGeneric` instead of `Refined` for `GroupName` (previously `Group.Name`).
- Use `ValidatedGeneric` instead of `Refined` for `GroupExpression`.
- Save `DeviceType` as a `String` in the DB instead of as `Int`s (and migrate the `Int`s).
- Bundle most Marshallers in a dedicated object.
- Rename and move codecs for consistency.
- Extract enumerations to its own files.
- Other minor tidying up.

### DIFF
--- a/src/main/resources/db/migration/V30__device_type_as_string.sql
+++ b/src/main/resources/db/migration/V30__device_type_as_string.sql
@@ -1,0 +1,5 @@
+ALTER TABLE Device
+MODIFY COLUMN device_type varchar(200) NOT NULL;
+
+UPDATE Device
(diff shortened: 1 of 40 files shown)


houcros:
Indeed this PR is not pressing, but any feedback here? It would be nice to start getting rid of `Refined` as we discussed. @koshelev @simao

houcros:
Replaced by #95.

PR #74, opened 2019-01-31
Add tracing support

simao:
Use libats-logging


houcros:
Nice!

houcros:
BTW how will this play with the Slick logs? It's defined in `logback.xml` right now.

simao:
You can change slick logging using `HIKARI_LOG_LEVEL` and `SLICK_LOG_LEVEL`. See https://github.com/advancedtelematic/libats/blob/fb5a596385579ded6a83eba7a72a2da5a37e495d/libats-logging/src/main/resources/logback.xml#L24

In any case, you can still create a local logback.xml with your changes, this will override the settings in `libats-logging`.

PR #75, opened 2019-02-05
Return report description in failure CSV export.

houcros:
Add missing failure description to the CSV export.

### DIFF
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
@@ -206,10 +206,10 @@ class DevicesResource(
     complete(db.run(action))
   }
 
-  def fetchFailureStats(correlationId: CorrelationId): Route = {
+  def fetchFailureStats(correlationId: CorrelationId, failureCode: Option[String]): Route = {
     implicit val exportMarshaller: Marshaller[Seq[(DeviceOemId, String, String)], HttpResponse] =
       Marshaller.oneOf(installationFailureMarshaller, installationFailureCsvMarshaller)
-    complete(db.run(InstallationReportRepository.fetchDeviceFailures(correlationId)))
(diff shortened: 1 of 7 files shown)


houcros:
With my last commit I had to add support to request the CSV with an extension in the request. The `Accept` header is still supported and preferred. But FE was having a hard time getting it to work like that, it seems the browser is doing some witchcraft with the headers.

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -223,8 +226,17 @@ class DevicesResource(
           case None      => complete(Errors.InvalidGroupExpression(""))
           case Some(exp) => countDynamicGroupCandidates(ns.namespace, exp)
         } ~
-        (path("failed-installations") & parameter('correlationId.as[CorrelationId])) {
-          cid => fetchFailureStats(cid)
+        (pathPrefix("failed-installations") & parameters('correlationId.as[CorrelationId], 'failureCode.as[String].?)) {
+          (cid, fc) => {
+            rawPathPrefix(".csv") {
+              mapRequest(addAcceptCsvHeader) {
```
simao: you don't need these headers... you might want to return a response with the `Content-Type` header, but not `Accept`, `Accept` is for the client, not the server.

houcros: Mmh but then I'm bypassing the content negotiation, and I thought that was the whole point...
If I can do that then I could return to the first implementation where this endpoint only returns CSV and sets the `Content-Type` header with a single Marshaller?

simao: yes, exactly if you use a .csv ending in the path and return csv in content-type you don't need to do any content negotiation, akka will handle that for you.

PR #82, opened 2019-03-20
[OTA-2308] Extend the CSV export endpoint to accept multiple correlationIds

houcros:
We need to update the CSV export to take into account the new retry-campaigns. Therefore the export endpoint now accepts multiple `correlationId`s to pass the main-campaign ID and the retry-campaign ID.


houcros:
I'm drafting an implementation in campaigner. Let's discuss both with Jerry this afternoon.

houcros:
Closing because we decided to implement this in campaigner after all. See https://github.com/advancedtelematic/campaigner/pull/111.

PR #83, opened 2019-03-22
[OTA-2308] Create endpoint to get DeviceOemIds from DeviceIds

houcros:
To be used when moving the CSV export to campaigner.

### DIFF
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
@@ -80,6 +80,7 @@ class DevicesResource(
   val eventJournal = new EventJournal()
 
   implicit val groupIdUnmarshaller: Unmarshaller[String, GroupId] = GroupId.unmarshaller
+  implicit val deviceIdUnmarshaller: Unmarshaller[String, DeviceId] = DeviceId.unmarshaller
 
   implicit val correlationIdUnmarshaller: FromStringUnmarshaller[CorrelationId] = Unmarshaller.strict {
     CorrelationId.fromString(_).leftMap(new IllegalArgumentException(_)).valueOr(throw _)
@@ -213,6 +214,16 @@ class DevicesResource(
     }
   }
 
+  def fetchDeviceOemIds(deviceIds: Iterable[DeviceId]): Route = {
+    val f = db.run {
+      DeviceRepository
+        .findByUuids(deviceIds.toSet)
+        .map(_.map(d => d.uuid -> d.deviceId))
+        .map(_.toMap)
+    }
+    complete(f)
+  }
+
   def api: Route = namespaceExtractor { ns =>
     val scope = Scopes.devices(ns)
     pathPrefix("devices") {
@@ -229,6 +240,9 @@ class DevicesResource(
         } ~
         (path("stats") & parameters('correlationId.as[CorrelationId], 'reportLevel.as[InstallationStatsLevel].?)) {
           (cid, reportLevel) => fetchInstallationStats(cid, reportLevel)
+        } ~
+        (path("oem-ids") & parameter('deviceId.as[DeviceId].*)) { deviceIds =>
+          fetchDeviceOemIds(deviceIds)
         } ~
         pathEnd {
           searchDevice(ns.namespace)
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
@@ -20,7 +20,7 @@ import akka.stream.ActorMaterializer
 import cats.syntax.either._
 import cats.syntax.show._
 import com.advancedtelematic.libats.auth.{AuthedNamespaceScope, Scopes}
-import com.advancedtelematic.libats.data.DataType.{CorrelationId, Namespace}
+import com.advancedtelematic.libats.data.DataType.{CorrelationId, DeviceOemId, Namespace}
 import com.advancedtelematic.libats.http.UUIDKeyAkka._
 import com.advancedtelematic.libats.messaging.MessageBusPublisher
 import com.advancedtelematic.libats.messaging_datatype.DataType.DeviceId._
@@ -32,7 +32,7 @@ import com.advancedtelematic.ota.deviceregistry.common.Errors
 import com.advancedtelematic.ota.deviceregistry.data.Codecs._
 import com.advancedtelematic.ota.deviceregistry.data.DataType.InstallationStatsLevel.InstallationStatsLevel
 import com.advancedtelematic.ota.deviceregistry.data.DataType.{DeviceT, InstallationStatsLevel, SearchParams, UpdateDevice}
-import com.advancedtelematic.ota.deviceregistry.data.Device.{ActiveDeviceCount, DeviceOemId}
+import com.advancedtelematic.ota.deviceregistry.data.Device.ActiveDeviceCount
 import com.advancedtelematic.ota.deviceregistry.data.Group.{GroupExpression, GroupId}
 import com.advancedtelematic.ota.deviceregistry.data.GroupType.GroupType
 import com.advancedtelematic.ota.deviceregistry.data.{CsvSerializer, PackageId}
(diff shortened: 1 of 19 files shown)


src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -213,6 +214,16 @@ class DevicesResource(
     }
   }
 
+  def fetchDeviceOemIds(ns: Namespace, deviceIds: Iterable[DeviceId]): Route = {
+    val f = db.run {
+      DBIO
+        .sequence(deviceIds.map(DeviceRepository.findByUuid))
```
nkly: Does it produce an sql query for each device UUID or somehow magically puts everything in one query? If it's the 1st case, I'd refactor it because it might hurt performance.

houcros: Well, I always assumed this was somehow using just one query. The docs are not very explicit: `Transform a TraversableOnce[ DBIO[R] ] into a DBIO[ TraversableOnce[R] ]`. I figured one `DBIOAction` implied one query. But running with debug logs it show indeed one query for each deviceId. (Or maybe it depends on the underlying DB?)
I'm not sure if this is big performance hit because we're using this method happily everywhere, and I'm also not aware if there is any way to do this in batch somehow. Maybe @simao?

nkly: One way to do in batch would be to filter using `_.inSet(deviceIds)` predicate

houcros: Indeed! I don't know why that didn't come to mind. Thanks, updated.

simao: DBAction encapsulated multiple statements into one db access using the pool, does not mean multiple statements and it doesn't even mean it needs to run on the same transaction.

`inSet` works here because I am guessing `deviceIds` will not be so big, but this would fail with a big enough set. 

Another way to batch this would be to split deviceIds into groups and build a DBAction for each group, then sequence that DBAction and run it with `transactionally`.

nkly: @simao Good point about the limitation of `inSet`. Do you know the exact number? I couldn't find the limit in MariaDB docs. If it's a reasonable number (100-1000), we can simply limit it on the endpoint and make clients do the batching.

simao: The limit will depend on the max query size, https://mariadb.com/kb/en/library/server-system-variables/#max_allowed_packet and https://mariadb.com/kb/en/library/query-limits-and-timeouts/

So way more than 100-1000.

houcros: I can't find exact requirements for this user story, but I'd say thousands of deviceIds should be more than enough. Given what I've seen I don't think a campaign with thousands of devices failing for the same reason is a frequent situation. Nonetheless, I could add a limit in this endpoint if we feel it's necessary.

nkly: @houcros @simao Well in this case I don't think we need to change anything because we'll notice HTTP delays faster than hit DB limit.

simao: That is not true for all services we have, but yes in this case.

simao:
Why is the csv export moving to campaigner? Why would campaigner need to know about OEM ids? Then we'd have that data in two places.

nkly:
@simao We're working on the retry campaign feature. One of the requirements for it is to allow user to export the list of OEM IDs of the devices that failed in CSV format, separately for each of the error codes. The list should be up-to-date, meaning that if there was a retry campaign that changed the state of devices (some suceeded, some failed with the same code, some failed with a different code), we need to adapt the list accordingly. 

After some discussions we've come up with 2 different paths: either we expose OEM IDs on the device registry (that is what's done in this PR) or we make it aware of retry campaigns somehow and move the logic of collecting and processing states of all the devices in all the main and retry campaigns to it. Though the 1st option is not ideal, as you've already said, I think that the 2nd approach is worse because it blurs the border between services, essentially moving a huge chunk of campaigner's job to the devices registry. 

Unfortunatly I'm not aware of any other viable approach to solve this task. If you have any ideas, feel free to comment.

simao:
> After some discussions we've come up with 2 different paths: either we expose OEM IDs on the device registry (that is what's done in this PR) or we make it aware of retry campaigns somehow and move the logic of collecting and processing states of all the devices in all the main and retry campaigns to it. Though the 1st option is not ideal, as you've already said, I think that the 2nd approach is worse because it blurs the border between services, essentially moving a huge chunk of campaigner's job to the devices registry.

We don't need to move the logic for collecting and processing states to device registry, you'd just need to move the result of that collection/processing, maybe put that data on the bus and let interested services consume it and allow export.

All this data should probably be in something like the auditor service. I don't think exposing this data so openly like this is a good idea, specially saving this oem id in campaigner, this data should not leak from device registry just like device name doesn't.

houcros:
Yeah I think we all agree this is ugly as sin. But sadly the other options we discussed seem no better 😕

simao:
Why can't we just move the processing result to device registry ?

houcros:
Campaigner is the only one that knows or can calculate the set of `DeviceId`s that need to be part of the export. If we could get those from here I guess we could. But then device-registry would need to know about campaigner, or maybe publish some message from campaigner and process it here. It's not clear that would be any better.

simao:
If campaigner publishes the result, registry would know if it should be included in the report or not. I think that would be better than leaking this data to other service.

houcros:
In that case, I think campaigner would need to publish messages every time a campaign or retry-campaign finishes to keep an up-to-date list of `DeviceId`s here. Either one message for campaign with a `Map[ResultCode, Seq[DeviceId]]` or one message for (campaign, resultCode) with a `Seq[DeviceId]`. We could store that in a new table here and keep it up-to-date removing or adding devices when we receive a new message for the same (campaign, resultCode). But could that turn out to be too many messages? (No idea.)

houcros:
My educated guess is it shouldn't be a problem: if we're able to process one message for each device that comes online, and it's hard to imagine more campaigns finishing at a given moment than devices coming online, this should be manageable as well...

simao:
yes, I don't think that should be a problem...

nkly:
> If campaigner publishes the result, registry would know if it should be included in the report or not. I think that would be better than leaking this data to other service.

@simao Could you please elaborate on this idea a little bit more? Right now I'm a bit puzzled about how the information from campaigner is going to be persisted and used, and also why is it better than sharing it the other way around?

houcros:
What about using the `/api/v1/devices/:deviceId` endpoint for each `deviceId` in campaigner to get the `oemId`s? Would it be _that_ horrible?

simao:
It's almost the same, that data is leaving this service.

houcros:
But that is the case now anyways, isn't it? Anyone can call that endpoint and get all the info about the device...

simao:
Yes, but we are not storing it anywhere else. I don't think the export should be in campaigner at all.

houcros:
Yes, we also don't mean to store it in campaigner. Just to fetch it (even if as a String) from dev-reg and spit it out in the CSV. If we keep the export here we're gonna need to listen to some campaigner messages and keep them in a new table to avoid using an endpoint that is already available. It's not obvious to me that is preferable 😕
Also, it seems to me that conceptually the endpoint should be under the `campaign` resource, since we're exporting devices that failed for a campaign. Now it's under the `devices` resource and that always bugged me a bit.

houcros:
So this is what we had in mind for the export: https://github.com/advancedtelematic/campaigner/pull/111.

koshelev:
What if we move an export logic into web app? It can call both campaigner and device registry to compile CSV and serve a result to UI?

houcros:
So we're not introducing this endpoint and using the available `/devices/:uuid` instead. Let's revisit in the future if it affects performance.

PR #85, opened 2019-04-10
Make the installation reports migration from auditor optional

### LINKED ISSUE #84: ota-device-registry totally broken
Sorry for reporting here, but https://github.com/advancedtelematic/ota-device-registry/ isn't configured to allow bug reports. I just updated a test cluster from:

  advancedtelematic/device-registry:0.0.3-10-g883acfc

to:

 advancedtelematic/device-registry:0.3.0-11-g41be2f8

Its completed hosed my cluster. My devices show up as never seen online. If I create new devices, they can't talk to the service. Worse yet, nothing changes if I rollback. So it seems to completely corrupt my cluster.

houcros:
Fixes #84.
This is to support the OTA Community Edition. Since the auditor is not a public service, we need to offer the possibility to make the installation reports migration optional. Otherwise it will always fail for users of the Community Edition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ Management of device status and configuration information. This project is part 
 
 ## Building
 
+You will have to set the environment variable `AUDITOR_MIGRATE=false` before running this service as part of the Community Edition.
+
 `sbt docker:publishLocal` - to publish docker image locally.  
 
 ## Running tests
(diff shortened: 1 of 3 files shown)


doanac:
Thanks!

houcros:
Happy to help @doanac :)
If you now run the latest version of device-registry `0.3.0-12-ga1ea92b` with `AUDITOR_MIGRATE=false` that pesky error should go away. Ping me back if something goes wrong.

PR #87, opened 2019-05-03
Remove installation reports export endpoint

houcros:
We moved this endpoint to campaigner so we should remove it from here.


doanac:
I'm not a user of this specific endpoint. However, as someone who maintains a production instance of OTA Connect, seeing breakages introduced into versioned APIs scares me. The only way I can safely stay up-to-date with these projects is to eyeball every PR to make sure nothing I need is being altered.

My suggestion - don't maintain the baggage, I wouldn't want to either. However, add a new v$X version and creates releases in these projects about new APIs, their backward incompatible changes, and when the old API is scheduled to be removed ideally giving the consumer at least 1 release where the user can update and then work out the transition from v$X -> v$X+1.

nkly:
@doanac Thank you for the suggestion. While you're correct about this change being a breaking one, this specific API is internal and is not supposed to be used outside the whole system. That's why it could be simply removed, given that all the other components are being updated along with Device Registry. 

Now the larger problem is whether a component needs an update and to which version. Unfortunately, we don't have a way to communicate that information to OSS users yet. We plan to a list of versions of each component that are compatible with each other for each major production update. I hope it'll ease maintenance.

PR #90, opened 2019-05-21
[OTA-2445] Do not fail when deleting non-existent device

nkly:


src/test/scala/com/advancedtelematic/ota/deviceregistry/daemon/DeleteDeviceHandlerSpec.scala
```diff
@@ -0,0 +1,25 @@
...
+    extends FunSuite
+    with Matchers
+    with ScalaFutures
+    with DatabaseSpec {
+
+  import scala.concurrent.ExecutionContext.Implicits.global
+  val handler = new DeleteDeviceHandler()
+
+  test("OTA-2445: do not fail when deleting non-existent device") {
+    val msg = DeleteDeviceRequest(Namespace("whatever"), DeviceId(UUID.randomUUID))
```
houcros: IIRC there are generators for namespace and deviceId, why not use those instead?

nkly: Good point. I just for some reason completely forgot about them. Will fix.

PR #91, opened 2019-06-04
[OTA-2641] Keep some info about deleted device in a separate table

nkly:
Fixes [OTA-2641](https://saeljira.it.here.com/browse/OTA-2641)

### DIFF
--- a/src/main/resources/db/migration/V30__add_deleted_devices.sql
+++ b/src/main/resources/db/migration/V30__add_deleted_devices.sql
@@ -1,0 +1,8 @@
+CREATE TABLE DeletedDevice (
+    namespace CHAR(200) NOT NULL,
+    device_uuid char(36) NOT NULL,
+    device_id varchar(200) NOT NULL,
+    `created_at` datetime(3) NOT NULL DEFAULT current_timestamp(3),
+    PRIMARY KEY(namespace, device_uuid, device_id)
+);
+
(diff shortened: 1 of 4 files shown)


doanac:
Nice idea. I've recently had to add a hack into our deployment to deal with this:

 https://github.com/foundriesio/ota-api/commit/6a494d6d981c986c2e76938c360cd6981218cddb

One thing I see missing here is code to prevent a deleted device from being re-created. Is that something that could be added to this PR?

simao:
@doanac It seems like recreating a deleted device could be a useful use case... Why do you think it should not be allowed ?

doanac:
@simao - i apologize because I no longer have the details on me, but we hit an issue where someone added their deleted device back by accident and things "didn't work". I don't remember exactly what didn't work, but we just decided to block it from happening. Maybe the real "fix" is to understand what broke when it was re-added though.

src/main/resources/db/migration/V30__add_deleted_devices.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE DeletedDevice (
+    namespace CHAR(200) NOT NULL,
+    device_uuid char(36) NOT NULL,
```
houcros: My OCD is crying on seeing this lower-case `char` mingled with all the upper-cases 🙈

src/main/resources/db/migration/V30__add_deleted_devices.sql
```diff
@@ -0,0 +1,7 @@
...
+    device_id varchar(200) NOT NULL,
```
houcros: I think we could benefit from a `created_at` column here, to know when the device was deleted.

src/main/resources/db/migration/V30__add_deleted_devices.sql
nkly: Apparently that's the convention used in this service :neutral_face:

PR #92, opened 2019-06-04
Do not use `ptest` suffix when creating test databases

simao:


deploy/ci_setup.sh
```diff
@@ -5,8 +5,6 @@ set -u
 mkdir entrypoint.d/ || true
 
 echo "
-CREATE DATABASE device_registry;
-CREATE DATABASE device_registry_test;
 GRANT ALL PRIVILEGES ON \`device\_registry%\`.* TO 'device_registry'@'%';
```
alexhumphreys: https://hub.docker.com/_/mariadb/ grep that page for the env var `MYSQL_USER`. Looks like that user already gets permissions on the db that is created with `MYSQL_DATABASE`, so you can maybe drop this whole section?

simao: this grant privilieges to all dbs *starting with* `device_registry`, not just `device_registry`

PR #95, opened 2019-06-17
Stop using `Refined`

houcros:
I remember we discussed getting rid of the `Refined` library, and that's why we implemented our `ValidatedGeneric` in libats. (Unfortunately I don't remember exactly why, but @koshelev might be able to fill that gap.) So, if this is still something worth doing, we can gradually do this for the various services and eventually remove `Refined` as a dependency in libats.

To be merged and deployed with https://github.com/advancedtelematic/ota-plus-server/pull/1997.

### DIFF
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
@@ -38,8 +38,6 @@ import com.advancedtelematic.ota.deviceregistry.data.GroupType.GroupType
 import com.advancedtelematic.ota.deviceregistry.data.{GroupExpression, PackageId}
 import com.advancedtelematic.ota.deviceregistry.db._
 import com.advancedtelematic.ota.deviceregistry.messages.{DeleteDeviceRequest, DeviceCreated}
-import eu.timepit.refined.api.Refined
-import eu.timepit.refined.string.Regex
 import io.circe.Json
 import slick.jdbc.MySQLProfile.api._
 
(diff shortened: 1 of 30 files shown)


src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -153,8 +150,8 @@ class DevicesResource(
     complete(db.run(InstalledPackages.getDevicesCount(pkg, ns)))
 
   def listPackagesOnDevice(device: DeviceId): Route =
-    parameters(('regex.as[String Refined Regex].?, 'offset.as[Long].?, 'limit.as[Long].?)) { (regex, offset, limit) =>
```
houcros: I'm waiting for FE confirmation, but it seems no one uses this regex parameter and we can just remove it.

houcros: Ok so it is used, but regex is overkill. I will replace for "string contains".

houcros: Done.

simao: but we need to change ui?

houcros: Yes, I'll make sure to sync with the FE before merging this.

houcros: Created parallel PR in app: https://github.com/advancedtelematic/ota-plus-server/pull/1997.

houcros:
Because it's a transitive dependency that we get from libats. The idea would be to stop using this in whatever services use it and then eventually remove it from libats.

zabbal:
> Because it's a transitive dependency that we get from libats. 

I'm still puzzled. If we have "import eu.timepit.refined.api.Refined" in our .scala than it looks like an explicit dependency. What am I missing?

houcros:
Yes, but if you see there is no declaration of the library in the `build.sbt` in this repo. There is, nonetheless, a declaration to include libats, and that in turn includes `refined` as a dependency here: https://github.com/advancedtelematic/libats/blob/cc903491991c2376fe840d421db5fe82593fec8e/build.sbt#L68.

PR #100, opened 2019-07-08
[OTA-3179] Correctly parse a CSV exported from a spreadsheet

houcros:
The current implementation fails because exporting as CSV from a spreadsheet application doesn't add a line break after the last record.


houcros:
The build fails for now because we need to merge this first: https://github.com/advancedtelematic/libats/pull/112

nkly:
Maybe it would be better to split Akka bumping and CSV fixing into separate PRs?

houcros:
Fair enough, will do.

houcros:
Opened #101.

PR #101, opened 2019-07-08
Bump libats and AkkaHttp versions

houcros:


houcros:
#100 is based on this changes and has been reviewed and approved, hence this can be considered as reviewed transitively. It's in different PRs just to keep things tidy.

PR #108, opened 2019-09-02
Only build docker image on push to master

simao:
This means that `git push deploy/sit` will not work if the revision
was not previously pushed to master, as the image to deploy would not
have been built yet. IMO this is a good thing as only revisions in
master can be deployed to SIT, but it might not be obvious when the
deployment fails.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,7 +40,6 @@ build docker image:
   stage: build image
   only:
     - master
-    - /^deploy\/.*/
   tags:
     - docker-sock
   variables:


alexhumphreys:
If we're spinning up environments on PR, won't they need to have images built?

simao:
Yes, but then they should be built as part of that pipeline?

alexhumphreys:
Maybe? I still don't know where that pipeline is, or how it'll be shared between the different repos, so I guess the answer depends on that.

simao:
I am not too sure about this PR to be honest

alexhumphreys:
As you like, it's a pretty easy change to revert :)

PR #110, opened 2019-09-05
Create endpoint to fetch never seen devices.

houcros:
Needed for the new UI homepage.
I could return a `PaginatedResult` but I'm not sure it is necessary or worth it here, WDYT?


src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -209,6 +209,9 @@ class DevicesResource(
         (path("stats") & parameters('correlationId.as[CorrelationId], 'reportLevel.as[InstallationStatsLevel].?)) {
           (cid, reportLevel) => fetchInstallationStats(cid, reportLevel)
         } ~
+        path("never_seen") {
```
koshelev: What if we create an `GET /devices?neverSeen=true` with pagination instead? AFAIR pagination result contains total number of objects

houcros: Sure, that's reasonable. Just one thing: I see in the model "never seen" is equivalent to "not activated". Would `GET /devices?activated=false` be more idiomatic? (To avoid double negation.)

houcros: Ah I remembered why I was aiming to make this endpoint at a different path: `/devices` already takes a combination of 7 different query parameters, with a PR open to add an 8th. I figured it was blowing a bit out of control with a 9th...

houcros: Updated to use API as `/devices?activated=<true|false>`.

PR #113, opened 2019-09-09
CI/CD: clarify stages/jobs naming

zabbal:


simao:
When do you want this to run?

zabbal:
> When do you want this to run?

What do you mean by "this"?

simao:
The pipeline

zabbal:
There're 3 pipelines touched by this patch. Two of those could only be triggered via schedule. One can only be triggered if particular variable is defined. So far this variable been defined as part of the schedule as well.

simao:
There is one pipeline to which you are adding 3 stages, when do you want those stages to run?

zabbal:
Ah, right, there're actually 2 pipelines spread over 3 stages. Nevertheless, it's all meant to run via schedule.

simao:
Where is the 2nd pipeline?

zabbal:
pipeline 1: "start deps scan" -> "generate pages"
pipeline 2: "start veracode scan"

simao:
There is only one pipeline that contains different stages, which contains different jobs that are either activated or not depending on the conditions we put on them

PR #115, opened 2019-09-10
Fix gitlab pages generation

zabbal:
Revert previously requested renaming as it breaks pages generation.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -116,7 +116,7 @@ start deps scan:
     paths:
       - depchk.html
 
-generate pages:
+pages:
   stage: publish dependencies scan
   only:
     - schedules


Raigi:
I guess you should rebase first: This branch is out-of-date with the base branch

simao:
Please rebase, not merge.

.gitlab-ci.yml
```diff
@@ -116,8 +116,8 @@ start deps scan:
     paths:
       - depchk.html
 
-generate pages:
-  stage: publish dependencies scan
+pages:
+  stage: pages
```
simao: Please use a better name for this, this is not the cause of this error, we use descriptive names for other jobs, see for example deployment descriptors.

alexhumphreys: https://docs.gitlab.com/ce/ci/yaml/README.html#pages I think the _stage_ can be renamed, but not the _job_.

simao: Ah! Ok, lets change the stage name then.

PR #116, opened 2019-09-17
Fix flaky test.

houcros:
This silly test fails sometimes because we create too many test objects and retrieve too few.


jochenschneider:
TC failed. Is that a different test?

houcros:
Damn it. That's a flaky test but for another reason.

houcros:
Submitted a fix for the other flakiness reason.

PR #118, opened 2019-10-08
Feat/deploy hat

rdanitz:
auto-deploy environment to EKS cluster;
See https://main.gitlab.in.here.com/olp/edge/ota/testing/dynamic-hats for more infos

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -44,11 +44,13 @@ test:
 build docker image:
   stage: build image
   except:
-    - schedules
+    refs:
+      - schedules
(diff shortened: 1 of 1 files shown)


rdanitz:
This will pick up `deploy/…` branches; will this interfere with the current infrastructure? This would also deploy a `deploy-sit` namespace for example.

simao:
> This will pick up `deploy/…` branches; will this interfere with the current infrastructure? This would also deploy a `deploy-sit` namespace for example.

We could either use `deploy/hat/:hat-name` or use `except: deploy/sit`

PR #122, opened 2019-10-11
Bump versions, add exception file

jochenschneider:
Bumped a few versions and added exception file for the dependency checker

Signed-off-by: Jochen Schneider <j.schneider@here.com>


jochenschneider:
In a report like https://olp.pages.gitlab.in.here.com/edge/ota/connect/back-end/ota-device-registry/ there are buttons labelled "suppress" which generate an entry. The notes are from me.

zabbal:
> In a report like https://olp.pages.gitlab.in.here.com/edge/ota/connect/back-end/ota-device-registry/ there are buttons labelled "suppress" which generate an entry. The notes are from me.

Neat. Might make sense to add this to commit message to make it easier to update this next time.

PR #123, opened 2019-10-15
Add number of devices to the 'list groups' endpoint.

houcros:
Wouldn't say this is very elegant, but it's needed for OTA-3667 (dashboard page of the new UI). In the "recently created" view we want to show how many devices there are in the groups.


src/main/scala/com/advancedtelematic/ota/deviceregistry/db/GroupInfoRepository.scala
```diff
@@ -44,11 +44,35 @@ object GroupInfoRepository {
...
+      .groupBy(_._2)
+      .map { case (g, q) => g -> q.length }
+
+    val groupsWithoutDevices = GroupMemberRepository.groupMembers
+      .joinRight(groups)
+      .on(_.groupId === _.id)
+      .filter(_._1.isEmpty)
+      .map { case (_, g) => g -> 0 }
+
+    groupsWithDevices
```
houcros: I wasn't able to do it with a single outer join, so I went with two joins and a union. If it's possible and simpler to do it with one join I'd gladly update this.

houcros:
Talked with @koshelev and we should better calculate this from app with the existing APIs.

PR #128, opened 2019-10-24
Publish known `system_info` data from device to message bus

simao:
When hardware information from the device the device through the
`/system_info` endpoint, make a best effort to parse the `product`
from the json and publish a message to the bus containing that data.

Currently we already publish a msg with that information, this commit
augments that message with an optional `product` attribute.


simao:
Depends on https://github.com/advancedtelematic/libats/pull/129

src/test/scala/com/advancedtelematic/ota/deviceregistry/SystemInfoResourceSpec.scala
```diff
@@ -158,4 +164,49 @@ class SystemInfoResourceSpec extends ResourcePropSpec {
...
+        |""".stripMargin
+    ).right.get
+
+    createSystemInfo(uuid, json) ~> route ~> check {
+      status shouldBe Created
+    }
+
+    eventually {
+      val msg = messageBus.findReceived[DeviceSystemInfoChanged](uuid.toString)
+      msg.flatMap(_.newSystemInfo).flatMap(_.product) should contain("test-product")
```
houcros: It could be nice to use ScalaTest's `OptionValues` here? Although I don't think it's such a big improvement... http://doc.scalatest.org/3.0.1/#org.scalatest.OptionValues

simao: Nice one, didn't know about it, I like it

src/test/scala/com/advancedtelematic/ota/deviceregistry/ResourceSpec.scala
```diff
@@ -12,7 +12,7 @@ import akka.http.scaladsl.server.Route
 import akka.http.scaladsl.testkit.{RouteTestTimeout, ScalatestRouteTest}
 import com.advancedtelematic.libats.auth.NamespaceDirectives
 import com.advancedtelematic.libats.data.DataType.Namespace
-import com.advancedtelematic.libats.messaging.MessageBus
+import com.advancedtelematic.libats.messaging.{MessageBus, MessageBusPublisher}
```
jochenschneider: Is this needed?

PR #132, opened 2019-11-13
Implement first version of public API

simao:
This is implemented as a separate resource and routes as it would be easier to extract and verify on code reviews.

We keep this in this in device-registry, rather than create a new service for this, so that we can provide an api to the users as soon as possible without going through setting up deployment descriptors and ota-charts. Since delivering something to the client by the end of the current sprint is a hard requirement from product, we have no choice here.

Unfortunately this means that device-registry now contacts director. We must be careful to not use this code from anywhere else in device-registry.

We are including `libtuf-server` so we can reuse at least some of the data type to parse data coming from director.

charts need to be updated to include `DIRECTOR_HOST` and `DIRECTOR_PORT`, or device registry will not startup anymore.

The requirements for OTA-3977 mention only:

    For a given device id,
    1. Device status (UpToDate, Error, NotSeen, Outdated)

Which this change implements. However, the final user already requested implementation of:

          2. Pending Update and its associated status (Download
          started, Download completed, Installation started,
          Installation completed)

          3. Current installed OSTree image on primary (hash)

So this change also implements a first version for 3.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,6 +24,7 @@ lazy val `ota-device-registry` =
       libraryDependencies += "org.tpolecat" %% "atto-core" % "0.7.1"
     )
     .settings(libraryDependencies ++= library.libAts)
+    .settings(libraryDependencies += "com.advancedtelematic" %% "libtuf-server" % "0.7.0-59-gf6013d6")
     .settings(dependencyOverrides += library.kafkaClient)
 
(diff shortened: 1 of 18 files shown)


simao:
ah yeah :+1:

simao:
Changed to v1alpha, added basic open api spec + swagger ui

PR #133, opened 2019-11-18
Use clientDeviceId instead of oemId

simao:
Fix http(s) urls for unpkg

### DIFF
--- a/src/main/resources/api-definition.yml
+++ b/src/main/resources/api-definition.yml
@@ -21,7 +21,13 @@ info:
     curl -H 'Authorization: Bearer $token' https://this-api-url/v1alpha/devices
     ```
 
-    The token will expire every 24 hours. The user needs to request a new token before the current token expires to continue using the api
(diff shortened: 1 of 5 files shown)


src/main/resources/api-definition.yml
```diff
@@ -21,7 +21,13 @@ info:
     curl -H 'Authorization: Bearer $token' https://this-api-url/v1alpha/devices
     ```
 
-    The token will expire every 24 hours. The user needs to request a new token before the current token expires to continue using the api
+    The token will expire every 24 hours. The user needs to request a new token before the current token expires to continue using the api.
```
jochenschneider: This could be misunderstood as "If you don't get a new token before the old one expires, you are screwed (because you need a valid token to get a new one for the next 24h)."
Better: "If the current token is expired, the user needs to request a new token to continue using the API."

simao: updated

PR #135, opened 2019-11-19
Use different json format for device

simao:
Use uuid instead of DeviceId

Use deviceId instead of clientDeviceId

Use installedSoftwareVersion instead of installedTarget

Remove `target` key from `installedSoftwareVersion`

### DIFF
--- a/src/main/resources/api-definition.yml
+++ b/src/main/resources/api-definition.yml
@@ -57,7 +57,7 @@ paths:
       description: Paginated list of devices
       parameters:
         - in: query
-          name: clientDeviceId
+          name: deviceId
           description: the id used by the client to provision the device
           schema:
             type: string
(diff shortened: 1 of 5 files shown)


src/main/resources/api-definition.yml
```diff
@@ -157,32 +157,30 @@ components:
           format: date-time
         primaryEcu:
           type: object
-          required: [ecuId, installedTarget]
+          required: [ecuId, installedSofwareVersion]
```
jochenschneider:
```suggestion
          required: [ecuId, installedSoftwareVersion]
```

src/main/resources/api-definition.yml
```diff
@@ -157,32 +157,30 @@ components:
...
           properties:
             ecuId:
               type: string
               format: string
-            installedTarget:
+            installedSofwareVersion:
```
jochenschneider: same.

src/main/resources/api-definition.yml
simao: Something wrong with github. I did have those wrong but I forced pushed a while ago with those fixed, don't know why doesn't update, if you see the commit it's fixed though.

PR #138, opened 2019-11-20
Do not run tests on deploy

simao:
Slightly controversial change. 

Tests are executed anyway when the image is built. If the image was not built (and the tests green) the deployment will fail, so why run the tests again?

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -20,6 +20,9 @@ test:
   stage: test
   tags:
     - docker-sock
+  except:
+    refs:
+      - deploy/sit
   services:
     - name: mariadb:10.2
       alias: db


jochenschneider:
For me that only makes sense if you could ensure that deploy/* only contains a ref which has been build and tested before.

(Different question: why is deploy/sit excluded for builds? https://github.com/advancedtelematic/ota-device-registry/blob/f92a08ca1c42a65c168c89f8ebd1d1ec43169102/.gitlab-ci.yml#L52)

simao:
> For me that only makes sense if you could ensure that deploy/* only contains a ref which has been build and tested before.
> 

You can guarantee.. if there is a build it was tested... 

what is the point of running tests again?

> (Different question: why is deploy/sit excluded for builds?
> 
> https://github.com/advancedtelematic/ota-device-registry/blob/f92a08ca1c42a65c168c89f8ebd1d1ec43169102/.gitlab-ci.yml#L52
> )

Why build an image again if already exists? it's actually a bad thing because we would be overwriting an already pushed image.

jochenschneider:
Ok, makes sense. What other "deploy/*" branches are being used other than "sit"?

simao:
No idea, I think some hat env deploys

jochenschneider:
What is the reason for excluding more than SIT from the tests?

simao:
Yeah, good point, build is only except on deploy/sit. Changed to only skip tests on sit.

PR #142, opened 2019-11-29
Fix the plural in the Device schema

xaviergully:
Feedback from Spekulatius : remove plural from hashes as this is not an array, but a single value

### DIFF
--- a/src/main/resources/api-definition.yml
+++ b/src/main/resources/api-definition.yml
@@ -164,14 +164,14 @@ components:
               format: string
             installedSoftwareVersion:
               type: object
-              required: [filename, length, hashes]
+              required: [filename, length, hash]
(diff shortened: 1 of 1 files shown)


simao:
It's a single hash for them, but might not be for others.
It's an object that can contain multiple hashes with different hashing algorithms. In their case only has one algorithm.

xaviergully:
Thanks @simao for your clarification.
In which cases there could be multiple hashes ?

simao:
It's one key in that object per hash algorithm, if we add sha512 to reposerver+director for example director would have a new key here.

simao:
hashes usage clarified here: https://github.com/advancedtelematic/ota-device-registry/pull/144

PR #144, opened 2019-12-02
Implement device queue and device events endpoints for public api

simao:
Implement device queue and device events endpoints for public api
    
    `devices/:device-uuid/queue` returns the queue for a device
    
    `devices/:device-uuid/events` returns events received from a device

Fixed a bug when getting events by correlation id

Correct a few api definition typos

### DIFF
--- a/src/main/resources/api-definition.yml
+++ b/src/main/resources/api-definition.yml
@@ -1,4 +1,4 @@
-openapi: 3.0.0
+qopenapi: 3.0.0
 info:
   description: |
     HERE OTA API for device information
@@ -105,6 +105,58 @@ paths:
         401:
           $ref: '#/components/responses/UnauthorizedError'
 
+
+  /devices/{uuid}/queue:
+    get:
+      tags:
+        - devices
+      description: Show the update queue for a device
+      parameters:
+        - name: uuid
+          in: path
+          description: ID of the device to return the queue
+          required: true
+          schema:
+            type: string
+            format: uuid
+      responses:
+        200:
+          description: Queued updates for the device
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/QueueItem"
+
+  /devices/{uuid}/events:
+    get:
+      tags:
+        - devices
+      description: Show all events reported by a device, sorted by most recent first
+      parameters:
+        - name: uuid
+          in: path
+          description: ID of the device
+          required: true
+          schema:
+            type: string
+            format: uuid
+        - name: updateId
+          in: query
+          description: Only return events that refer to this update
+          required: false
+          schema:
+            type: string
+            format: string
+      responses:
+        200:
+          description: A description of the current status of the update in this device
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/DeviceEvents"
+
 components:
   securitySchemes:
     bearerAuth:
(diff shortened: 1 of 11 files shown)


src/main/resources/api-definition.yml
```diff
@@ -139,6 +194,82 @@ components:
...
+            type: object
+            required: [ecuId, name, timestamp]
+            properties:
+              ecuId:
+                type: string
+                format: string
+              name:
+                type: string
+                enum:
+                  - DownloadComplete
```
xaviergully: Do all of these states exist on the portal ? 
I am just aware of downloadStart, downloadCompleted, installationStart, installationCompleted

simao:
A previous version of this PR implemented a different API, the one discussed during planning:

/devices/:device-uuid/queue
/devices/:device-uuid/queue/:update-id/status

However, this is not possible to implement reliably using the data we have in our system. We don't have such a thing as "update status", so anything we would return to the client would be derived from the device events. But this would not be reliable, it's subject to change depending on the events the device pushes and the way device registry parses them, and it's not directly related to an update. Those events can, for example, not include a correlationId, which would mean we have no way to map that to an update.

For this reason, we implement a different API here:

```
/devices/:device-uuid/queue
/devices/:device-uuid/events?updateId=<updateId> # `updateId` is optional and is available in the `/queue` response
```

`/events` returns all events for a device. Optionally filtered for events that refer to an `updateId`. They are sorted by `deviceTime`, the timestamp the device attributed to the event. This is a similar API used by our UI, which then infers an update state, and this is what API clients will need to do. The last event in `/events` is the last known event for that update on that device and the api users will know what happened to the update last by checking the name of the event.

src/main/resources/api-definition.yml
simao: yes they exist.

simao:
@Raigi new endpoints docs available here: http://api.porto.sit-ota.aws.in.here.com/docs You could start writing tests based on that schema if you are blocked

I will deploy to sit soon if @houcros and @jochenschneider don't raise any issues.

src/main/scala/com/advancedtelematic/ota/api_provider/http/ApiProvider.scala
```diff
@@ -46,4 +46,47 @@ class ApiProvider(directorClient: DirectorClient)(implicit ec: ExecutionContext,
...
+        }
+
+      }.toVector.flatten
+    }
+  }
+
+  def findUpdateEvents(namespace: Namespace, deviceId: DeviceId, correlationId: Option[CorrelationId]): Future[ApiDeviceEvents] = async {
+    val indexedEvents = await(eventJournal.getIndexedEvents(deviceId, correlationId))
+
+    println(indexedEvents.map(_._2.eventType))
```
jochenschneider: Leftover?

src/main/scala/com/advancedtelematic/ota/deviceregistry/DevicesResource.scala
```diff
@@ -254,6 +255,8 @@ class DevicesResource(
         path("events") {
           import DevicesResource.EventPayloadDecoder
           (get & parameter('correlationId.as[CorrelationId].?)) { correlationId =>
+            // TODO: This should not return raw Events
+            // API should not return arbitrary json (`payload`) to the clients. This si why we index interesting events, so we can give this info to clients
```
jochenschneider:
```suggestion
            // API should not return arbitrary json (`payload`) to the clients. This is why we index interesting events, so we can give this info to clients
```

jochenschneider: Do you want to create a ticket for this?

src/test/scala/com/advancedtelematic/ota/api_provider/http/DeviceInfoResourceSpec.scala
```diff
@@ -116,4 +125,136 @@ class DeviceInfoResourceSpec extends FunSuite with ResourceSpec with Eventually
...
+      "correlationId" -> campaignId01.toString().asJson
+    ).asJson
+
+    val payload02 = Map(
+      "ecu" -> ecuId.asJson,
+      "correlationId" -> campaignId02.toString().asJson
+    ).asJson
+
+    val event01 = Event(deviceId, UUID.randomUUID().toString, EventType("EcuInstallationStarted", 0), now, now, payload01)
+    new DeviceEventListener().apply(DeviceEventMessage(defaultNs, event01)).futureValue
```
jochenschneider: Put the DeviceEventListener into a variable and call it twice?

PR #145, opened 2019-12-03
Aktualizr config processing

jochenschneider:
Signed-off-by: Jochen Schneider <j.schneider@here.com>

### DIFF
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/SystemInfoResource.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/SystemInfoResource.scala
@@ -8,29 +8,37 @@
 
 package com.advancedtelematic.ota.deviceregistry
 
+import java.time.Instant
+
 import akka.Done
 import akka.actor.ActorSystem
+import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.model.StatusCodes._
 import akka.http.scaladsl.server.{Directive1, Route}
 import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.unmarshalling.{FromEntityUnmarshaller, Unmarshaller}
 import com.advancedtelematic.libats.auth.{AuthedNamespaceScope, Scopes}
 import com.advancedtelematic.libats.data.DataType.Namespace
 import com.advancedtelematic.libats.messaging.MessageBusPublisher
-import com.advancedtelematic.ota.deviceregistry.common.Errors.MissingSystemInfo
+import com.advancedtelematic.ota.deviceregistry.common.Errors.{Codes, MissingSystemInfo}
 import com.advancedtelematic.ota.deviceregistry.db.SystemInfoRepository
 import com.advancedtelematic.ota.deviceregistry.db.SystemInfoRepository.NetworkInfo
 import io.circe.{Decoder, Encoder, Json}
-import org.slf4j.LoggerFactory
 import slick.jdbc.MySQLProfile.api._
 import com.advancedtelematic.libats.messaging_datatype.DataType.DeviceId
 import com.advancedtelematic.libats.http.UUIDKeyAkka._
 import cats.syntax.show._
 import cats.syntax.option._
-import com.advancedtelematic.libats.messaging_datatype.Messages
-import com.advancedtelematic.libats.messaging_datatype.Messages.{DeviceSystemInfoChanged, SystemInfo}
+import com.advancedtelematic.libats.http.Errors.RawError
+import com.advancedtelematic.libats.messaging_datatype.Messages.{AktualizrConfigChanged, DeviceSystemInfoChanged}
+import toml.Toml
+import http.`application/toml`
 
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.ExecutionContext
 
+case class AktualizrConfig(uptane: Uptane, pacman: Pacman)
+case class Uptane(polling_sec: Int, force_install_completion: Boolean)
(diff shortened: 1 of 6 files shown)


src/main/scala/com/advancedtelematic/ota/deviceregistry/SystemInfoResource.scala
```diff
@@ -60,6 +62,12 @@ class SystemInfoResource(
     } yield (uuid: DeviceId) => NetworkInfo(uuid, ip, hostname, mac)
   }
 
+  implicit val aktualizrConfigUnmarshaller: FromEntityUnmarshaller[AktualizrConfig] = Unmarshaller.stringUnmarshaller.map { s =>
+    import toml.Codecs._
+
+    Toml.parseAs[AktualizrConfig](s).right.get
```
simao: IIRC this unmarshaller api allows you to return a future, or an option, something like that, when this fails, would be better than just failing like this?

src/main/scala/com/advancedtelematic/ota/deviceregistry/SystemInfoResource.scala
```diff
@@ -126,6 +134,17 @@ class SystemInfoResource(
                 }
               complete(NoContent -> result)
             }
+          } ~
+          path("config") {
+            pathEnd {
+              post {
+                entity(as[AktualizrConfig]) { config =>
+                  val result = messageBus.publish(AktualizrConfigChanged(ns.namespace, uuid, config.uptane.polling_sec,
```
simao: `result` is just a unit IIRC, we should just return 202/Accepted or 204/NoContent with no content at all.

src/main/scala/com/advancedtelematic/ota/deviceregistry/SystemInfoResource.scala
simao: This generates a codec with macros or how is this working?

doanac:
This is interesting. I assume a change is about to go into Aktualizr to send this data?

src/main/scala/com/advancedtelematic/ota/deviceregistry/SystemInfoResource.scala
jochenschneider: It uses Shapeless which I guess uses macros? Similar to Circe, I imagine, though I also don't really know how that works.

simao: But is the codec defined somewhere or is just this?

jochenschneider: I am using "import toml.Codecs._" which derives codecs. You can also define your own.

simao:
@doanac yes, that will follow

src/main/scala/com/advancedtelematic/ota/deviceregistry/SystemInfoResource.scala
jochenschneider: > IIRC this unmarshaller api allows you to return a future, or an option, something like that, when this fails, would be better than just failing like this?

To me it looks like that Unmarshallers can't do more than throw an exception in that case. I can convert it to a more generic one.

simao: yes but would be better to let unmarshallers handle that case that to just call `.get` here.

src/main/scala/com/advancedtelematic/ota/deviceregistry/SystemInfoResource.scala
jochenschneider: `result` contains a `Future[Done]`. What this does is that it returns no content in the good case and a 500 if the future failed, which seems ok to me.

simao: Returning 500 when it fails is fine, but this is returning `NoContent` with a response. Done marshalls an empty string. it's returning an empty string as a content. No Content should be no content.

jochenschneider: I thought this was an idiom, it is used a few lines above. What is the difference between “no content” and an empty string? The content type? `entity.isKnownEmpty` holds.

simao: I don't understand the question

NoContent is just for the status code, it's not for the actual response entity. It is an idiom if you return Ok -> entity, then returns 200 with some entity, but if you return NoContent it should not have an entity at all.

jochenschneider: Ah, right. My question was if it does actually return an entity. What is the difference between an empty entity and no entity in the HTTP response?

simao: By empty I meant like an empty string which is what I think `Done` marshalls to. like `""`. This would mean the content-type will be `application/json` and some clients will try to parse this and fail. 

No response is no response at all and this would also mean no `content-type` header

if you use `val f = result.map(_ => NoContent); complete(f)` this would return 204 with no entity at all

jochenschneider: Ah yeah that's shorter than what I just did. I did see that the content type is "text" during the tests.

simao:
sorry, pushed to the wrong branch. should be fixed now.

PR #152, opened 2019-12-06
Use event names from IndexedEventType for public API

simao:
Initially I thought it would be a good idea to simplify these events
and remove `Ecu` and `Campaign` from them, but this just confuses
users further. So here I am removing that translation and using
IndexedEventType directly.

Signed-off-by: Simão Mata <simao.mata@here.com>

### DIFF
--- a/src/main/resources/api-definition.yml
+++ b/src/main/resources/api-definition.yml
@@ -213,16 +213,16 @@ components:
                 type: string
                 enum:
                   - DownloadComplete
-                  - DownloadStarted
-                  - DownloadCompleted
-                  - InstallationStarted
-                  - InstallationApplied
-                  - InstallationCompleted
+                  - EcuDownloadStarted
+                  - EcuDownloadCompleted
+                  - EcuInstallationStarted
+                  - EcuInstallationApplied
+                  - EcuInstallationCompleted
                   - DevicePaused
                   - DeviceResumed
-                  - Accepted
-                  - Declined
-                  - Postponed
+                  - CampaignAccepted
+                  - CampaignDeclined
+                  - CampaignPostponed
                   - InstallationComplete
               receivedTime:
                 type: string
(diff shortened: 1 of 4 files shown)


src/main/resources/api-definition.yml
```diff
@@ -213,16 +213,16 @@ components:
                 type: string
                 enum:
                   - DownloadComplete
```
Raigi: Other events changed to EcuDownload.., but what for is this one? Something special?

src/main/resources/api-definition.yml
```diff
@@ -213,16 +213,16 @@ components:
...
+                  - EcuInstallationCompleted
                   - DevicePaused
                   - DeviceResumed
-                  - Accepted
-                  - Declined
-                  - Postponed
+                  - CampaignAccepted
+                  - CampaignDeclined
+                  - CampaignPostponed
                   - InstallationComplete
```
Raigi: Same question for this event

src/main/resources/api-definition.yml
simao: We should ask the embedded team, this comes from the client. I think this might be if the device has multiple ecus, it sends this after all ecus are installed? @patrickvacek  ?

patrickvacek: The events that libaktualizr sends can be seen in https://github.com/advancedtelematic/aktualizr/blob/master/src/libaktualizr/primary/reportqueue.cc. I don't see a `DownloadComplete`.

src/main/resources/api-definition.yml
```diff
@@ -213,16 +213,16 @@ components:
...
+                  - EcuInstallationApplied
+                  - EcuInstallationCompleted
                   - DevicePaused
                   - DeviceResumed
-                  - Accepted
-                  - Declined
-                  - Postponed
+                  - CampaignAccepted
+                  - CampaignDeclined
+                  - CampaignPostponed
```
patrickvacek: Does it matter that the libaktualizr versions of the Campaign events appear to be in snake_case instead of CamelCase? @lbonn do you recall why that was?

simao: we receive it in snake case but converted it here to camel case for the public api

PR #153, opened 2019-12-09
Do not return duplicate queue items

simao:
One of the many bugs of director is that it sometimes returns
duplicate queue items, even if they refer to the same update, on the
same ecu. This change makes sure we only return one of those updates

### DIFF
--- a/src/test/scala/com/advancedtelematic/ota/api_provider/http/DeviceInfoResourceSpec.scala
+++ b/src/test/scala/com/advancedtelematic/ota/api_provider/http/DeviceInfoResourceSpec.scala
@@ -26,6 +26,7 @@ import com.advancedtelematic.ota.deviceregistry.daemon.DeviceEventListener
 import com.advancedtelematic.ota.deviceregistry.data.DataType.IndexedEventType
 import io.circe.syntax._
 import org.scalatest.OptionValues._
+import org.scalatest.LoneElement._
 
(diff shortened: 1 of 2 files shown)


src/test/scala/com/advancedtelematic/ota/api_provider/http/DeviceInfoResourceSpec.scala
```diff
@@ -260,4 +260,24 @@ class DeviceInfoResourceSpec extends FunSuite with ResourceSpec with Eventually
...
+    val hash = "848cba347e8a37330b97835936dd4f846291739d0d5efa9eb10c75e4c15ba87a".refineTry[ValidChecksum].get
+    val image = EcuInfoImage(targetFilename, 2222, Hashes(hash))
+
+    directorClient.addQueueItem(deviceId, DirectorQueueItem(correlationId.some, Map(ecuId -> image)))
+    directorClient.addQueueItem(deviceId, DirectorQueueItem(correlationId.some, Map(ecuId -> image)))
+
+    Get(apiProviderUri("devices", deviceId.show, "queue")) ~> route ~> check {
+      status shouldBe StatusCodes.OK
+      val queue = responseAs[List[QueueItem]]
+      queue should have size(1)
```
houcros: You could do `responseAs[List[QueueItem]].loneElement shouldBe queueItem` if you mix in `org.scalatest.LoneElement`, but that's just being fancy 😄

jochenschneider: wow

PR #154, opened 2019-12-11
Add more description to the Events list

xaviergully:
This is just a proposal from my understanding
I am not sure about the formatting as well.
Feel free to amend it

### DIFF
--- a/src/main/resources/api-definition.yml
+++ b/src/main/resources/api-definition.yml
@@ -110,6 +110,7 @@ paths:
     get:
       tags:
         - devices
+      summary: Queue for a device
       description: Show the update queue for a device
       parameters:
         - name: uuid
@@ -133,6 +134,7 @@ paths:
     get:
       tags:
         - devices
+      summary: Events reported by a device
       description: Show all events reported by a device, sorted by most recent first
       parameters:
         - name: uuid
@@ -211,19 +213,18 @@ components:
                 format: string
               name:
                 type: string
-                enum:
-                  - DownloadComplete
-                  - EcuDownloadStarted
-                  - EcuDownloadCompleted
-                  - EcuInstallationStarted
-                  - EcuInstallationApplied
-                  - EcuInstallationCompleted
-                  - DevicePaused
-                  - DeviceResumed
-                  - CampaignAccepted
-                  - CampaignDeclined
-                  - CampaignPostponed
-                  - InstallationComplete
+                enum: [EcuDownloadStarted, EcuDownloadCompleted, EcuInstallationStarted, EcuInstallationApplied, EcuInstallationCompleted, DevicePaused, DeviceResumed, CampaignAccepted, CampaignDeclined, CampaignPostponed]
+                description:  >
+                  EcuDownloadStarted - Package download for this ECU has started <br>
+                  EcuDownloadCompleted - Package download for this ECU is completed <br>
+                  EcuInstallationStarted - Package installation for this ECU has started <br>
+                  EcuInstallationApplied - Package has been applied to this ECU but is pending reboot <br>
+                  EcuInstallationCompleted - Package installation for this ECU is completed <br>
+                  DevicePaused - The device has been paused (downloads are paused and API calls are deferred) <br>
+                  DeviceResumed - The device has resumed after being paused <br>
+                  CampaignAccepted - Campaign has been accepted by the user or it has been automatically accepted by the device <br>
+                  CampaignDeclined - Campaign has been declined by the user <br>
+                  CampaignPostponed - Campaign has been postponed by the user <br>
               receivedTime:
                 type: string
                 format: date-time


xaviergully:
@tkfu, I am not able to add Amma as a reviewer

src/main/resources/api-definition.yml
```diff
@@ -212,18 +212,18 @@ components:
...
-                  - InstallationComplete
+                  - DownloadComplete : Download complete for the whole campaign ?
+                  - EcuDownloadStarted : Package download for this ECU has started  
+                  - EcuDownloadCompleted : Package download for this ECU is completed
+                  - EcuInstallationStarted : Package installation for this ECU has started
+                  - EcuInstallationApplied : ?
+                  - EcuInstallationCompleted : Package installation for this ECU is completed
+                  - DevicePaused : The device has seen set on pause, i.e. all libaktualizr API call are queued ?
+                  - DeviceResumed : The device has resumed, i.e. all libaktualizr API call are processed ?
+                  - CampaignAccepted : Campaign has been accepted by the user or it has been automatically accepted by the device
```
Raigi: not shown when campaign is automatically accepted

src/main/resources/api-definition.yml
```diff
@@ -212,18 +212,18 @@ components:
...
-                  - DeviceResumed
-                  - CampaignAccepted
-                  - CampaignDeclined
-                  - CampaignPostponed
-                  - InstallationComplete
+                  - DownloadComplete : Download complete for the whole campaign ?
+                  - EcuDownloadStarted : Package download for this ECU has started  
+                  - EcuDownloadCompleted : Package download for this ECU is completed
+                  - EcuInstallationStarted : Package installation for this ECU has started
+                  - EcuInstallationApplied : ?
```
Raigi: Saw this one when Installing packages on a primary that had secondaries

simao:
Does not compile:

```
Parser error 
bad indentation of a sequence entry
Jump to line 219
 ```

**Comment by simao** [comment]
See https://swagger.io/docs/specification/data-models/enums/

File: src/main/resources/api-definition.yml [hunk 3]
```diff
@@ -211,19 +211,21 @@ components:
...
-                  - CampaignDeclined
-                  - CampaignPostponed
-                  - InstallationComplete
+                enum: [DownloadComplete, EcuDownloadStarted, EcuDownloadCompleted, EcuInstallationStarted, EcuInstallationApplied, EcuInstallationCompleted, DevicePaused, DeviceResumed, CampaignAccepted, CampaignDeclined, CampaignPostponed, InstallationComplete]
+                description:  >
+                  DownloadComplete - Download complete for the whole campaign <br>
+                  EcuDownloadStarted - Package download for this ECU has started <br>
+                  EcuDownloadCompleted - Package download for this ECU is completed <br>
+                  EcuInstallationStarted - Package installation for this ECU has started <br>
+                  EcuInstallationApplied - ?? <br>
```
jochenschneider: What is the deal with this?

src/main/resources/api-definition.yml
```diff
@@ -211,19 +211,21 @@ components:
...
+                  EcuInstallationCompleted - Package installation for this ECU is completed <br>
+                  EcuInstallationCompleted - Package installation for this ECU is completed <br>
+                  DevicePaused - The device has seen set on pause, i.e. all libaktualizr API call are queued ? <br>
```
jochenschneider: Another question mark

src/main/resources/api-definition.yml
```diff
@@ -211,19 +211,21 @@ components:
...
+                  DeviceResumed - The device has resumed, i.e. all libaktualizr API call are processed ? <br>
```
jochenschneider: Another question mark

src/main/resources/api-definition.yml
```diff
@@ -211,19 +211,21 @@ components:
...
+                  CampaignAccepted - Campaign has been accepted by the user or it has been automatically accepted by the device <br>
+                  CampaignDeclined - Campaign has been declined by the user <br>
+                  CampaignPostponed - ? <br>
```
jochenschneider: Another question mark

src/main/resources/api-definition.yml
```diff
@@ -211,19 +211,21 @@ components:
...
+                  InstallationComplete - Installation complete for the whole campaign ? <br>
```
jochenschneider: Another question mark

src/main/resources/api-definition.yml
xaviergully: Question mark means that I don't know, so difficult for me to explain or describe it. Anyone would have a suggestion ?

src/main/resources/api-definition.yml
xaviergully: Question mark means that I don't know, so difficult for me to explain or describe it. Anyone would have a suggestion ?

src/main/resources/api-definition.yml
xaviergully: Question mark means that I don't know, so difficult for me to explain or describe it. Anyone would have a suggestion ?

src/main/resources/api-definition.yml
xaviergully: Question mark means that I don't know, so difficult for me to explain or describe it. Anyone would have a suggestion ?

src/main/resources/api-definition.yml
jochenschneider: That's what I figured, just doesn't seem ideal to me to "document" something like that. Even leaving it out would be better IMO.

jochenschneider: I imagine the embedded people know these events well?

patrickvacek: True. I'll push a commit to this PR if that works.

patrickvacek: Look better now?

By the way, I looked into `DownloadComplete` and it was deprecated a while back. I also tried to look for references to `InstallationComplete` but it looks like libaktualizr never sent it. It seems that was never actually implemented, and has anyway since become deprecated by the `EcuInstallation` events.

jochenschneider: To me it looks much better now, thanks!

patrickvacek: Cool. It sounds like you can probably get rid of `InstallationComplete` entirely; no need for backwards compatibility if we literally never used it. Is it actually used on the BE?

xaviergully: Looks better now. 
Why not getting rid of `DownloadComplete` and `InstallationComplete` if both are deprecated ?

patrickvacek: > Why not getting rid of `DownloadComplete` and `InstallationComplete` if both are deprecated ?

At least for `DownloadComplete`, backwards compatibility with users still using a version from the brief period where we used that. Or if that isn't important, I have no objection to getting rid of it.

xaviergully: which version of aktualizr are we talking about ? How old would it be ?

patrickvacek: added in 47bb4a5d656ccfb6b1cef6f35515aba03ccfca5a (June 2018), superseded in 1aca27723c0d8912cd33b4a76fc004de7542cd86 (October 2018), completely removed in ff0cb488d2654413f89f59b20e21d9f9715d7d9d (October 2018).

xaviergully: I checked which version of aktualizr are currently out there on active devices from our main users, customers : most of them are using a version more recent than October 2018. So I think it is safe to get rid of both

Raigi:
I think you need to rebase

jochenschneider:
You merged master into this branch, which creates a merge commit, would be nicer if you could rebase instead. (Ask me if you don't know what we mean.)

houcros:
While at it, squashing all the commits would be nice 😃

xaviergully:
Thanks @jochenschneider

simao:
We are still parsing DownloadComplete and InstallationComplete. This PR only changes documentation, it doesn't change the behavior if these events somehow end up in the api.

xaviergully:
@simao I don't expect any clients sending the events DownloadComplete and InstallationComplete, because most of the customers are not using the aktuliazr version which sends these two events.

Do you want to keep the parsing for a while for monitoring ? And then it could be removed ?

patrickvacek:
Has anyone ever seen an instance of `InstallationComplete` sent by libaktualizr? In my digging I could not find a time such an event was ever sent.

Raigi:
> Has anyone ever seen an instance of `InstallationComplete` sent by libaktualizr? In my digging I could not find a time such an event was ever sent.

I have not

PR #159, opened 2020-02-06
[OTA-3654] Migrate blacklisting feature from ota-core

houcros:
The now-deprecated ota-core service serves a "blacklisting" feature. The quotes are because it's not actually blacklisting anything. It is used to count the number of devices that have a particular package installed. This are the installed packages reported by aktualizr, e.g. 'nano' for a linux distribution, not to be confused with the images we can install through our system.

We move that functionality here to discontinue the ota-core service.

The actual migration is in a stand-alone commit that can (and probably should) be reverted once the blacklist has been migrated in all the environments.

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -54,6 +54,14 @@ server {
   port = ${?BIND_PORT}
 }
 
+sota_core = {
+  host = "localhost"
+  host = ${?SOTA_CORE_HOST}
+  port = 8080
+  port = ${?SOTA_CORE_PORT}
+  uri = "http://"${sota_core.host}":"${sota_core.port}
+}
+
 main {
   defaultNs = "default"
   defaultNs = ${?DEFAULT_NAMESPACE}
--- a/src/main/resources/db/migration/V31__add_package_list_items.sql
+++ b/src/main/resources/db/migration/V31__add_package_list_items.sql
@@ -1,0 +1,10 @@
+CREATE TABLE PackageListItem (
+  namespace VARCHAR(255) NOT NULL,
+  package_name VARCHAR(200) NOT NULL,
+  package_version VARCHAR(200) NOT NULL,
+  `comment` TEXT NOT NULL,
+  created_at DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
+  updated_at DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3),
+
+  PRIMARY KEY (namespace, package_name, package_version)
+);
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/DeviceRegistryRoutes.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/DeviceRegistryRoutes.scala
@@ -29,6 +29,7 @@ class DeviceRegistryRoutes(
             new DevicesResource(namespaceExtractor, messageBus, deviceNamespaceAuthorizer).route ~
             new SystemInfoResource(messageBus, namespaceExtractor, deviceNamespaceAuthorizer).route ~
             new PublicCredentialsResource(namespaceExtractor, messageBus, deviceNamespaceAuthorizer).route ~
+            new PackageListsResource(namespaceExtractor, deviceNamespaceAuthorizer).route ~
             new GroupsResource(namespaceExtractor, deviceNamespaceAuthorizer).route
           }
         }
--- a/src/main/scala/com/advancedtelematic/ota/deviceregistry/PackageListsResource.scala
+++ b/src/main/scala/com/advancedtelematic/ota/deviceregistry/PackageListsResource.scala
@@ -1,0 +1,83 @@
+package com.advancedtelematic.ota.deviceregistry
+
+import akka.http.scaladsl.model.StatusCodes.{Created, NoContent}
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server._
+import com.advancedtelematic.libats.auth.{AuthedNamespaceScope, Scopes}
+import com.advancedtelematic.libats.data.DataType.Namespace
+import com.advancedtelematic.libats.messaging_datatype.DataType.DeviceId
+import com.advancedtelematic.ota.deviceregistry.data.Codecs._
+import com.advancedtelematic.ota.deviceregistry.data.DataType.{PackageListItem, PackageListItemCount}
+import com.advancedtelematic.ota.deviceregistry.data.PackageId
+import com.advancedtelematic.ota.deviceregistry.db.PackageListItemRepository
+import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport._
+import io.circe.generic.auto._
+import slick.jdbc.MySQLProfile.api._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+  * This "package lists" feature has been migrated from the deprecated
+  * ota-core service, where it used to be a "blacklisting" feature. It's
+  * been migrated here to terminate ota-core.
+  *
+  * The feature was not actually blacklisting anything. Instead, it was
+  * used to count the number of devices that have a particular package
+  * installed. This are the installed packages reported by aktualizr,
+  * e.g. 'nano' for a linux distribution, not to be confused with the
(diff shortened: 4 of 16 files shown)


houcros:
Tested the migration in `berlin` with 1 blacklisted package in one namespace and 4 in another namespace. Test OK.

src/main/resources/application.conf
```diff
@@ -54,6 +54,14 @@ server {
...
 }
 
+sota_core = {
+  host = "localhost"
+  host = ${?SOTA_CORE_HOST}
+  port = 8080
+  port = ${?SOTA_CORE_PORT}
+  uri = "http://"${sota_core.host}":"${sota_core.port}
+}
+
```
simao: you'll need to be careful running this migration.

Currently migrations in the gitlab pipelines run with the configmap of the previous version. This is because the configmap changes are only applied by helm when you helm install and we only helm install after migration because we only want to deploy after helm migration. So you'll need to first deploy the version with the new chart version and vars and then deploy the docker container, we cannot deploy docker+charts at the same time.

src/main/resources/db/migration/V31__add_blacklisted_packages.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE BlacklistedPackage (
+  namespace VARCHAR(255) NOT NULL,
+  package_name VARCHAR(200) NOT NULL,
+  package_version VARCHAR(200) NOT NULL,
+  `comment` TEXT NOT NULL,
+  created_at DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
+  updated_at DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3),
+
+  PRIMARY KEY (namespace, package_name, package_version)
+);
```
simao: Should we use a different name for all this other than blacklist? And just keep the api endpoints as legacy?

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/BlacklistedPackageRepository.scala
```diff
@@ -0,0 +1,70 @@
...
+import com.advancedtelematic.ota.deviceregistry.data.DataType.{BlacklistedPackage, BlacklistedPackageCount}
+import com.advancedtelematic.ota.deviceregistry.data.PackageId
+import com.advancedtelematic.ota.deviceregistry.db.DeviceRepository.devices
+import com.advancedtelematic.ota.deviceregistry.db.InstalledPackages.installedPackages
+import com.advancedtelematic.ota.deviceregistry.db.SlickMappings._
+import slick.jdbc.MySQLProfile.api._
+
+import scala.concurrent.ExecutionContext
+
+object BlacklistedPackageRepository {
```
simao: This came directly like this from core? I can't remember show the rest of device-registry is setup, but this repository is return DBIO and not Future, so it leaks that db logic a lot. We should at least keep it consistent with the rest of dev registry.

src/main/resources/application.conf
jochenschneider: Interesting. I copied the previous paragraph to the Gitops page on the Wiki.

jochenschneider:
I think it would be good to change the name from "blacklist" and comment in the main class what the feature actually does if it is not obvious from the name.

houcros:
Regarding the naming: yes, I agree we could have some better naming. But to be honest, after thinking for a while I couldn't come up with anything that makes sense to me... This is a very particular feature, after all...
I did write some explanation in the commits, but I think it also makes sense to have a short explanation in the class that serves the blacklist routes, will do that.

src/main/resources/application.conf
houcros: Thanks, didn't know that, will keep it in mind.

src/main/resources/db/migration/V31__add_blacklisted_packages.sql
houcros: Please see my other comment. I'm open to improvements, but I couldn't think of anything sensible myself...

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/BlacklistedPackageRepository.scala
houcros: Thing is, most of the other `*Registry` classes also return `DBIO[_]`, so in that sense it is consistent with the rest of the service...

src/main/resources/db/migration/V31__add_blacklisted_packages.sql
houcros: There is already an `InstalledPackages` table... `MarkedPackages`? `CountedPackages`? Don't know, it's all equally confusing for me... But at least `BlacklistedPackage` can somehow be traced back to its original intention 🤷‍♂️

src/main/scala/com/advancedtelematic/ota/deviceregistry/db/BlacklistedPackageRepository.scala
simao: yes that is what I meant by keeping it consistent, lets keep it like this then.

src/main/resources/db/migration/V31__add_blacklisted_packages.sql
simao: Could be something like "package list"? And then the concept of some package being in a list? And for now "One user => one list?".

houcros: Mmh I think it'd be "one namespace => one list". But apart from that I guess it makes some sense, yes. At least it's just a list and not a blacklist. Will push a fixup.

houcros:
See last commit for renaming from `BlacklistedPackage` to `ListedPackage`. At least it's a plain list and not a blacklist 😄
I went with `ListedPackage` instead of something like `PackageList`, because that would kind of imply that each record in the new table is a list of packages, when actually each record is one element of a list of packages.
Will squash that commit before merging.

src/main/scala/com/advancedtelematic/ota/deviceregistry/ListedPackagesResource.scala
```diff
@@ -0,0 +1,83 @@
...
+import com.advancedtelematic.ota.deviceregistry.data.PackageId
+import com.advancedtelematic.ota.deviceregistry.db.ListedPackageRepository
+import de.heikoseeberger.akkahttpcirce.FailFastCirceSupport._
+import io.circe.generic.auto._
+import slick.jdbc.MySQLProfile.api._
+
+import scala.concurrent.{ExecutionContext, Future}
+
+/**
+  * This "listed packages" feature has been migrated from the deprecated
```
jochenschneider: Nice!

simao:
> See last commit for renaming from `BlacklistedPackage` to `ListedPackage`. At least it's a plain list and not a blacklist smile
> I went with `ListedPackage` instead of something like `PackageList`, because that would kind of imply that each record in the new table is a list of packages, when actually each record is one element of a list of packages.
> Will squash that commit before merging.

I don't know, I think `ListedPackage` could work, but `PackageList` or `PackageLists`, or `PackageListItem` also works, because it's a common way to model a list within a table, using weak entities. See https://en.wikipedia.org/wiki/Weak_entity

houcros:
> > See last commit for renaming from `BlacklistedPackage` to `ListedPackage`. At least it's a plain list and not a blacklist smile
> > I went with `ListedPackage` instead of something like `PackageList`, because that would kind of imply that each record in the new table is a list of packages, when actually each record is one element of a list of packages.
> > Will squash that commit before merging.
> 
> I don't know, I think `ListedPackage` could work, but `PackageList` or `PackageLists`, or `PackageListItem` also works, because it's a common way to model a list within a table, using weak entities. See https://en.wikipedia.org/wiki/Weak_entity

Mmh ok, having seen that it seems it makes sense. I think I prefer `PackageListItem`, I'll go with that and add some comment in case it's not clear in the future.

houcros:
Amended last commit renaming to `PackageListItem`. Notice I didn't rename the migration classes because I will remove that code once it run, and because I think in that case it actually makes more sense to use the legacy naming, since it's connecting to the legacy service.
Will squash that commit later.

simao:
lgtm

houcros:
Tested one last time in `berlin`. Looking good, so let's do it 🚀

PR #164, opened 2020-02-17
Change of id parameter to deviceUuid

ammaboateng:
Within the schemas "Device" and "SimplifiedDevice", I changed the name `id` to `deviceUuid` in order to maintain consistency with the other schema descriptions of the `uuid`.

Signed-off-by: amma <awuraaamma.okyere-boateng@here.com>

### DIFF
--- a/src/main/resources/api-definition.yml
+++ b/src/main/resources/api-definition.yml
@@ -172,7 +172,7 @@ components:
       - uuid
       - deviceId
       properties:
-        id:
+        deviceUuid:
           type: string
           format: uuid
         deviceId:
@@ -274,14 +274,14 @@ components:
 
     Device:
       type: object
-      required: [uuid, deviceId, name, status]
+      required: [deviceId, uuid, name, status]
       properties:
-        id:
-          type: string
-          format: uuid
         deviceId:
           type: string
           format: string
+        deviceUuid:
+          type: string
+          format: uuid
         name:
           type: string
           format: string
--- a/src/main/resources/api-definition.yml
+++ b/src/main/resources/api-definition.yml
@@ -274,7 +274,7 @@ components:
 
     Device:
       type: object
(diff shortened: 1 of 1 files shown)


src/main/resources/api-definition.yml
```diff
@@ -274,14 +274,14 @@ components:
 
     Device:
       type: object
-      required: [uuid, deviceId, name, status]
+      required: [deviceId, uuid, name, status]
```
jochenschneider:
```suggestion
      required: [deviceId, deviceUuid, name, status]
```

ammaboateng: @jochenschneider I only changed the name and not the format of `uuid`. Changing the format here from `uuid` to `deviceUuid` would mean changing it for the other schemas not only in the documentation but also in the code. Because when the API is applied to a device the output is displayed as `uuid` and not `deviceUuid`.

jochenschneider: The original "required" list contained "uuid" which didn't exist and corresponds to "id". Since you renamed "id" to "deviceUuid" in the properties I suggest renaming "uuid" to "deviceUuid" in the "required" list.

ammaboateng: @jochenschneider ok I will rename it. Should I do the same to other areas with `uuid` parameter mentioned or does it only apply to this section?

jochenschneider:
Ahaha, good point! The file is now in this repo: https://main.gitlab.in.here.com/olp/edge/ota/connect/back-end/api-provider

jochenschneider:
My fault, sorry. Created a MR to remove the file here.

xaviergully:
@ammaboateng , please open a MR in https://main.gitlab.in.here.com/olp/edge/ota/connect/back-end/api-provider/blob/master/src/main/resources/api-definition.yml

